PR #5, opened 2016-02-15
Add datepicker

ihadeed:
Should I go as deep as documenting each sub-parameter for the parameter `options` ?

### DIFF
--- a/src/plugins/datepicker.ts
+++ b/src/plugins/datepicker.ts
@@ -1,0 +1,60 @@
+import {Plugin, Cordova} from './plugin';
+
+/**
+ * The DatePicker plugin allows the user to fetch date or time using native dialogs.
+ *
+ * Platforms supported: iOS, Android, Windows
+ *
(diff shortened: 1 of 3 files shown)


src/plugins/datepicker.ts
```diff
@@ -0,0 +1,60 @@
...
+ *
+ * @usage
+ * ```js
+ * DatePicker.
+ * ```
+ *
+ */
+@Plugin({
+  name: 'DatePicker',
+  plugin: '',
```
tlancina: I believe `plugin` is 'cordova-plugin-datepicker' and `pluginRef` is 'plugins.datePicker'

src/plugins/datepicker.ts
```diff
@@ -0,0 +1,60 @@
...
+  pluginRef: ''
+})
+export class DatePicker {
+
+  /**
+   * Shows the date and/or time picker dialog(s)
+   * @param options
+   * @returns {Promise<Date>} Returns a promise that resolves with the picked date and/or time, or rejects with an error.
+   */
+  static show(options? : {
```
tlancina: Needs the `@Cordova()` decorator

ihadeed:
Fixed.

PR #20, opened 2016-03-06
add batterystatus

ihadeed:
Not complete yet, thought I would show you where I'm going with this first. Note that I'm new to observables so I do apologize if there are any mistakes.

I thought for this plugin using the Cordova wrapper would not make sense since the original plugin does not have any functions to be called. 

*Edit*: I probably have to call isInstalled function, but I'm not sure what the pluginRef is for this plugin ...


tlancina:
Yeah I like this approach, looks fine to me :+1:

PR #40, opened 2016-03-10
feat(plugin): add imagePicker plugin

cho-is:


src/plugins/imagepicker.ts
```diff
@@ -0,0 +1,61 @@
...
+ *   for (var i = 0; i < results.length; i++) {
+ *       console.log('Image URI: ' + results[i]);
+ *   }
+ * }, (err) => {
+ * });
+ * ```
+ */
+@Plugin({
+  plugin: 'cordova-plugin-image-picker',
+  pluginRef: 'navigator.imagepicker',
```
cho-is: true!

Thanks for spotting it!

ihadeed: No problem :smile:

PR #41, opened 2016-03-10
Add tests

ihadeed:
@tlancina What do you think?


test/app/app.spec.js
```diff
@@ -0,0 +1,53 @@
...
+    IonicNative.Geolocation.getCurrentPosition();
+    expect(window.navigator.geolocation.getCurrentPosition).toHaveBeenCalled();
+  });
+
+  it('should watch position', function() {
+    spyOn(window.navigator.geolocation, 'watchPosition');
+    spyOn(window.navigator.geolocation, 'clearWatch');
+    var _geoLocationSubscription = IonicNative.Geolocation.watchPosition().subscribe();
+    expect(window.navigator.geolocation.watchPosition).toHaveBeenCalled();
+    _geoLocationSubscription.unsubscribe();
```
ihadeed: probably this should go inside another it();

tlancina:
Looking good to me :+1:

I wonder if there is a way to include the plugin scripts automatically from their npm package somehow, instead of needing to mock them.

ihadeed:
What about the plugins that require you to be on the device for them to
work!

PR #44, opened 2016-03-12
chore(docs): change ionic to cordova in commands

ihadeed:
Shouldn't we use `cordova` instead of `ionic` in case of non-IonicCLI users trying to use Ionic Native?


tlancina:
Yes, this should be Cordova. Looks good to me :+1:

Issue #46, opened 2016-03-13
Discussion: Adding @platforms parameter to docs


ihadeed:
Many cordova plugins work only on specific platforms. Some even have platform specific functions. 

It would be helpful to add a processor for `@platforms` parameter to docs generation process .. 

Example:
```js
/**
 * @name Base64 To Gallery
 * @description This plugin allows you to save base64 data as a png image into the device
 * @platforms Android, iOS, Windows Phone
 * @usage
 * ```ts
 * Base64ToGallery.base64ToGallery(base64Data, 'img_').then(
 *   res => console.log("Saved image to gallery ", res),
 *   err => console.log("Error saving image to gallery ", err)
 * );
 * ```
 */
```

Should output the following in docs:

# Base64 To Gallery
This plugin allows you to save base64 data as a png image into the device
### Usage
```ts
Base64ToGallery.base64ToGallery(base64Data, 'img_').then(
  res => console.log("Saved image to gallery ", res),
  err => console.log("Error saving image to gallery ", err)
);
```

PR #52, opened 2016-03-13
chore(docs): add platforms decorator

ihadeed:
https://github.com/driftyco/ionic-native/issues/46

Editor fixed the spacing ..

The things I added are in lines:

140 - 146
and
205 - 208


scripts/docs/templates/common.template.html
```diff
@@ -137,6 +137,14 @@ <h1 class="api-title">
...
 <@ endif -@>
 
+<!-- @platforms tag -->
+<@ if doc.platforms @>
+<h2>Supported platforms</h2>
+<@ block platforms @>
+<$ doc.platforms $>
+<@ endblock @>
+<@ endif @>
+
```
ihadeed: This will handle the plugin's main supported platforms

scripts/docs/templates/common.template.html
```diff
@@ -194,6 +202,11 @@ <h1 class="api-title">
 <div id="<$ method.name $>"></div>
 <h3><$ functionSyntax(method) $></h3>
 
+<@ if method.platforms @>
```
ihadeed: and this will take care of platform specific methods

tlancina:
What if we add this information to the `@Plugin` and `@Cordova` decorators and then possibly do runtime platform checks, since I believe Dgeni can also pull doc information out of the decorators.

PR #59, opened 2016-03-15
Show import format and link to ionic-native

textbook:
The plugin docs should mention ionic-native and the required import format, which we only found in https://forum.ionicframework.com/t/camera-plugin-not-working-build-error/45004/4?u=jonrsharpe after searching for quite a while.


ihadeed:
Thanks

PR #67, opened 2016-03-17
Toast 2.5.0 duration change

EddyVerbruggen:
The plugin now supports an arbitrary number of milliseconds for `duration`


ihadeed:
Thanks @EddyVerbruggen !

PR #80, opened 2016-03-24
docs(toast): Add import statement for Toast

wli:
Makes it clear how to import the Toast component into projects.


ihadeed:
Thanks!

PR #91, opened 2016-03-28
Connection interface doesn't need to be imported

ihadeed:
@keithdmoore


keithdmoore:
That needs to stay

keithdmoore:
@ihadeed So you can do this, for example:
```js
 if (Network.connection === Connection.WIFI) {
   console.log('we got a wifi connection, woohoo!');
 }
```

ihadeed:
Ok :+1:

PR #94, opened 2016-03-29
Added diagnostic plugin.

beachygreg:
I did not add every method and have not tested on IOS only android.


ihadeed:
Thanks for the PR @beachygreg !

Issue #86, opened 2016-03-26
Discuss: A good way to implement instance based plugins


ihadeed:
Many good plugins out there that we want to implement are instance based.

For example, look at the usage of the SQLite plugin:
```js
// first we create a database object
var db = window.sqlitePlugin.openDatabase(....);

// then we call it's methods
db.executeSql(...);
```

Now if we were to wrap this plugin and add it to Ionic Native, we wouldn't really be improving anything since we can't wrap the internal (prototype) methods. Not being able to wrap these internal methods means that we still have to use callback functions. Therefore, the Angular 2 change detection will not be triggered.

So the purpose of this discussion to gather ideas on how we can approach these type of plugins in a better way.

All thoughts and suggestions are welcome!

Issue #79, opened 2016-03-23
Cordova SQLite Plugin


muboch:
Will be nice to have this one too. 
Is this on the roadmap?
Link: https://github.com/litehelpers/Cordova-sqlite-storage.git

Thanks!

jdnichollsc:
is better this version (include pre-populated databases): https://github.com/litehelpers/cordova-sqlite-ext

PR #97, opened 2016-03-29
feat(ionicnative): add instance wrapper

ihadeed:
closes https://github.com/driftyco/ionic-native/issues/86
closes https://github.com/driftyco/ionic-native/issues/79

@tlancina Came up with this over night, it worked with the SQLite plugin. Haven't tested with other plugins. Let me know what you think.

Example of implementation: https://github.com/driftyco/ionic-native/blob/instance-wrapper/src/plugins/sqlite.ts


tlancina:
Nice! I'd maybe remove the console logs and have CordovaInstance call wrapInstance directly so you can get rid of `options.instanceMethod`, but other than that looks good.  We have nothing for this right now, so might as well get it in and test it out and see what the issues are.

keithdmoore:
@ihadeed Where are you at on this one?   I am looking to add keychain and will need to use something like this.  Should I use the instance_wrapper branch or are you pretty close to push these changes to master?

ihadeed:
Hey Keith,

If I remember correctly, the concept I had in mind is already implemented
but it hasn't been tested thoroughly.

I'm using the instance_wrapper branch to develop the wrapper and some of
the plugins that would rely on it.

I would suggest to develop on that branch so we can alter the instance
wrapper in case of any problems.
On Apr 18, 2016 11:44 AM, "Keith D. Moore" <notifications@github.com> wrote:

@ihadeed <https://github.com/ihadeed> Where are you at on this one? I am
looking to add keychain and will need to use something like this. Should I
use the instance_wrapper branch or are you pretty close to push these
changes to master?

—
You are receiving this because you were mentioned.

Reply to this email directly or view it on GitHub
<https://github.com/driftyco/ionic-native/pull/97#issuecomment-211439019>

Issue #12, opened 2016-02-23
Feature Request: Email Composer Plugin


danbucholtz:
I would love to see the email composer as part of Ionic Native.  We use it in several apps and I bet a lot of other companies do too.  Super common use case to build a CSV of some data and let it be shared via email.  This plugin works great for attachments.

http://ngcordova.com/docs/plugins/emailComposer/

Happy to help if you'll accept the PR.  Gotta read through the code to see what's going on.

Thanks,
Dan

tlancina:
Yeah would definitely love a PR. Check out https://github.com/driftyco/ionic-native/blob/master/DEVELOPER.md and https://github.com/driftyco/ionic-native/tree/master/src/plugins for some examples, and let me know if you have any questions/issues/feedback, thanks!

cmanou:
This can be closed since #110 was merged, and 1.2.0 can be released?

PR #110, opened 2016-04-05
feat(plugin): add email composer plugin

cmanou:
First pass at email composer plugin. Should address #12

### DIFF
--- a/src/plugins/emailcomposer.ts
+++ b/src/plugins/emailcomposer.ts
@@ -4,6 +4,7 @@ import {Plugin, Cordova} from './plugin';
  * Email object for Opening Email Composer
  */
 export interface email {
+  app?: string,
   to: string | Array<string>,
   cc: string | Array<string>,
   bcc: string | Array<string>,
@@ -52,11 +53,14 @@ export interface email {
 })
 export class EmailComposer {
   
-  @Cordova()
-  static isAvailable() : Promise<boolean> {return}
+  @Cordova({
+    successIndex: 1,
+    errorIndex: 3
+  })
+  static isAvailable(app: string, scope: any) : Promise<boolean> {return}
 
   @Cordova()
-  static addAlias(app: String, schema: any) : Promise<any> {return}
+  static addAlias(alias: string, packageName: string): void {}
   
   /**
    * Opens Email Composer with email contents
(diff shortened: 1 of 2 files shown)


src/plugins/emailcomposer.ts
```diff
@@ -0,0 +1,63 @@
...
+})
+export class EmailComposer {
+
+  /**
+   * Opens Email Composer with email contents
+   * @param email {email} Email
+   * @returns {Promise<any>} Resolves promise when the EmailComposer has been opened
+   */
+  @Cordova()
+  static open(email: email): Promise<any> { return }
```
ihadeed: Shouldn't it be something like this? 

```js
@Cordova({
successIndex: 1,
errorIndex: 3
})
static open(email: email, scope : any) : Promise<any> {return}
```

src/plugins/emailcomposer.ts
```diff
@@ -0,0 +1,63 @@
...
+ * ```
+ */
+@Plugin({
+  plugin: 'cordova-plugin-email-composer',
+  pluginRef: 'emailComposer',
+  repo: 'https://github.com/katzer/cordova-plugin-email-composer.git',
+  platforms: ['Android', 'iOS', 'Windows Phone 8']
+})
+export class EmailComposer {
+
```
ihadeed:
```js
@Cordova()
static isAvailable() : Promise<boolean> {return}
```

cmanou: Missed that when I looked at the ngCordova docs for the plugin will work on adding this. 
Thoughts on whether as first pass we need to implement the overload with the urischeme passed in as well?

cmanou:
Hi All, 
This is working for now though their is an issue with the isAvailable wrapper, since the root call in the cordova plugin the callback "moves" arguments depending on whether arguments are supplied, see [here](https://github.com/katzer/cordova-plugin-email-composer/blob/master/www/email_composer.js#L62)

I don't see an immediate way to do this with the current @Cordova() helper, unless i'm mistaken. What are your opinions on the best way to approach this?

ihadeed:
@cmanou yeah this plugin seems a bit tricky. 

We can wrap the whole function without the `@Cordova` wrapper, something like this:
```js
static isAvailable (app? : string) : Promise<boolean> {
return new Promise<boolean>((res, rej) => {
cordova.plugins.email.isAvailable(app, (isAvailable : boolean) => { if(isAvailable) res(); else rej() });
});
}
```

cmanou:
@ihadeed decided to go the no wrapper way though, kept it as resolving promise boolean as was the original plugin. Think this is now good to go?

ihadeed:
Thanks @cmanou !

Issue #112, opened 2016-04-06
Tests


rapropos:
What is the process for adding tests for modifications or new plugins? Does one simply add things to test/app.js? I have some changes and additions to the File plugin that I would like to consider contributing, but I want to make sure that things are working properly and consistently with what has been contributed so far.

ihadeed:
Hey Robert,

AFAIK there is no testing process as of now.  I did attempt to create one
at some point but it wasn't completed, as my approach had some problems
with it.

An ideal test would be to make sure that the Ionic native wrapper method is
invoking the plugin method.

When I test plugins, I usually load up a blank Ionic app on my Android
device that has all plugins installed. I make sure to include Ionic native
in there some how. Then I use chrome remote debug to access the plugins via
the console.

I would type in something like this in the console
`IonicNative.Dialogs.alert("....")`

PR #116, opened 2016-04-06
alternate file implementation

rapropos:
I guess this could be considered a work in progress, because I would like to be able to verify correct operation, but for your consideration here is an alternate implementation of  the File plugin that:

* includes TypeScript interface definitions and uses them in return values instead of `Promise<any>`
* contains all methods from the ngCordova plugin
* refactors the promise wrapping to hopefully have cleaner and more readable code without so many cascading error handlers

### DIFF
--- a/src/plugins/file.ts
+++ b/src/plugins/file.ts
@@ -1,6 +1,408 @@
-import {Plugin, Cordova} from './plugin';
+import {Plugin, Cordova} from "./plugin";
 
-declare var window;
+interface Window {
+  /**
+   * Requests a filesystem in which to store application data.
+   * @param type              Whether the filesystem requested should be persistent, as defined above. Use one of TEMPORARY or PERSISTENT.
+   * @param size              This is an indicator of how much storage space, in bytes, the application expects to need.
+   * @param successCallback   The callback that is called when the user agent provides a filesystem.
+   * @param errorCallback     A callback that is called when errors happen, or when the request to obtain the filesystem is denied.
+   */
+  requestFileSystem(type: number,
+                    size: number,
+                    successCallback: (fileSystem: FileSystem) => void,
+                    errorCallback?: (fileError: FileError) => void): void;
+  /**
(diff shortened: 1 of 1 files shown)


ihadeed:
Hey @rapropos,

Thanks for your input here.

You mentioned it is `work in progress`. Is that still the case or have you managed to get this to a stable level?

Also have you had a chance to look at the new instance wrapper which might be helpful for this plugin? 
See here for an example of implementation: https://github.com/driftyco/ionic-native/blob/master/src/plugins/media.ts

rapropos:
The "work in progress" comment was related to [my earlier question about tests](https://github.com/driftyco/ionic-native/issues/112), because I don't have an automated repeatable way to verify correct behavior. I think it works, but I can't prove it, and my app code doesn't exercise every single method. If you are still interested, I can look at incorporating the instance wrapper. I note that in that media plugin example all the return values are `Promise<any>`. Is it possible to use that wrapper with more specific `Promise<SpecificInterface>` return types? That is something that is important to me.

ihadeed:
@rapropos It would be great to implement the instance wrapper, yes. 

Also, it is possible to add custom interfaces for each return type. We are constantly working on adding more types. We publish plugins without types just to get them out there ASAP, even if they lack types and documentation.

rapropos:
@ihadeed I took another look at this, and don't see an easy way to integrate the instance wrapper here, because there is so much pre- and post-processing needed to be done on the JavaScript side of the bridge for most if not all publicly visible actions. I could split the typings added at the top of the file out into a file in the types/ directory if that would improve the chances of this getting accepted.

ihadeed:
Hey @rapropos I just tested some stuff out.

What you have there looks good but I think we will need to work on it more before publishing it. 

Could you please submit another PR to the `dev` branch that I just created? Thanks!

rapropos:
@ihadeed thank you for your attention. please see [#217](https://github.com/driftyco/ionic-native/pull/217).

PR #122, opened 2016-04-14
Append Trailing Slash If Not Present On path Param

Barryrowe:
This minor addition will automatically append a trailing slash ```/``` when the path param does not end with a slash and is used to build a directory path via concatenation.


ihadeed:
Thanks @Barryrowe , I will add this for now but we might be changing the whole implementation for this plugin at some point. Will probably reuse your fix if necessary.

PR #125, opened 2016-04-15
documentation wrong for watchAcceleration

koraysels:
documentation was wrong. watchPosition instead of watchAcceleration..


ihadeed:
Thanks @koraysels !

PR #132, opened 2016-04-18
Correct battery-status "status" event name

Rybbow:
As seen in the [plugin documentation](https://github.com/apache/cordova-plugin-battery-status#batterystatus-event) the event used should be called "batterystatus" instead of "batterylevel". Currently the observable you provide in the onChange method never fires, beacuse it's listening to a wrong event. This pull request aims to fix that.


ihadeed:
Thanks for the PR @Rybbow !

Issue #135, opened 2016-04-25
Badge plugin triggers exception in browser (using es6-shim)


wli:
I migrated my app over to use es6-shim (https://github.com/driftyco/ionic2-starter-tutorial/commit/ed9ef2fcce887e4d1c08c375c849b06b8394bad7), and these errors started triggering and I've narrowed it down to the Badge plugin.

```
EXCEPTION: Error: Uncaught (in promise): [object Object]
angular2-polyfills.js:528 Unhandled Promise rejection: Object {error: "cordova_not_available"} ; Zone: angular ; Task: Promise.then ; Value: Object {error: "cordova_not_available"}consoleError @ angular2-polyfills.js:528drainMicroTaskQueue @ angular2-polyfills.js:557ZoneTask.invoke @ angular2-polyfills.js:493
angular2-polyfills.js:530 Error: Uncaught (in promise): [object Object](…)
```

Can't repro in plunkr, since I can't figure how to add ionic-native to it, but the code is here: http://plnkr.co/edit/zBHmkYg7wAJpEdPRJ9f0?p=preview

```
this.platform.ready().then(() => {
      // This line is fine
      StatusBar.styleDefaut();
      
      // This line will crash
      Badge.clear();
    });
```

Here's a zip file of a fresh starter project with these changes in case it helps
[tutorial.zip](https://github.com/driftyco/ionic-native/files/235531/tutorial.zip)

wli:
And it looks like this is related to cordova.js being mocked out by `ionic serve`. This PR looks like it attempts to address this for the Badge plugin (which does support the browser platform):  https://github.com/driftyco/ionic-app-lib/pull/12

ihadeed:
Hey @wli 

The issue should be resolved in the next release (will be published soon).

wli:
Thanks for the fix @ihadeed! I'll be waiting for the release and can help confirm if this fixes the problem.

srishiva89:
Hey @ihadeed 

I'm getting error like "**Failed: Uncaught (in promise): cordova_not_available**" when running npm test command.

ihadeed:
Not sure what `npm test` command you're talking about @srishiva89

PR #139, opened 2016-04-27
fix(plugin): handle rejection when Cordova is undefined

tlancina:
Zone.js now (helpfully) throws an error on unhandled rejections in
promises. This is great because they no longer silently fail, but we
were relying on that behavior for the case where the plugin isn't
installed or Cordova isn't available. With this change, those cases
won't throw an unhandled rejection error, but all regular plugin errors
still should.


tlancina:
Addresses https://github.com/driftyco/ionic-native/issues/135

Issue #70, opened 2016-03-18
Cordova File Transfer Plugin


zackarychapple:
I have a new application that I am working on using Ionic2 with Typescript.  One of the requirements is offline video downloads.  I would like to be able to use Ionic-Native to perform the download of the videos and saving on the device.  Is this on the roadmap?

ihadeed:
Yes. This plugin will be added soon: https://github.com/apache/cordova-plugin-file-transfer

zackarychapple:
Fantastic.  I'm subscribed to this thread so hope when it lands it gets tagged.

diosney:
Hi, sorry if I'm trolling with this.

This is different as http://ngcordova.com/docs/plugins/file, right?

Is `cordova-plugin-file` already supported?

Thanks

diosney:
Hi, sorry if I'm trolling with this.

This is different as http://ngcordova.com/docs/plugins/file, right?

Is `cordova-plugin-file` already supported?

Thanks

ihadeed:
Yes that's correct. They are different. The `file` plugin is partially
implemented.

dtaalbers:
Keeping an eye on this issue! Can't wait until this will be added.

ghenry22:
Any news on the file transfer plugin?

kewix:
Would be very useful to mee too !

ihadeed:
The plugin is added and is available in the current NPM package and Github release. Sorry I forgot to close this earlier!

aneeskodappana:
the transfer plugin doesn't ask for permission on android. 
isn't fine to use this.file.dataDirectory instead of cordova.file.dataDirecotory as per the ionic wrapper documentation.

[Transfer plugin docs ](https://ionicframework.com/docs/native/transfer/)

```
download() {
  	const transfer: TransferObject = this.transfer.create();
  	const url = encodeURI("http://someimageurl.com/file.png");

  	transfer.download(url, this.file.dataDirectory + "file.png")
  		.then((entry) => {
  			alert("File downloaded successfully " + entry.toURL());
  		}, (error) => {
  			alert(JSON.stringify(error));
  		});
  }
```

the above code doesn't work. the app also not asking fo storage permission. what is the problem here?

PR #151, opened 2016-05-02
Implemented wrappers for FileTransfer.

glecaros:
Added wrappers for the file transfer plugin.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -31,6 +31,7 @@ import {Dialogs} from './plugins/dialogs';
 import {EmailComposer} from './plugins/emailcomposer';
 import {Facebook} from './plugins/facebook';
 import {File} from './plugins/file';
+import {Transfer} from './plugins/filetransfer';
 import {Flashlight} from './plugins/flashlight';
 import {Geolocation} from './plugins/geolocation';
 import {Globalization} from './plugins/globalization';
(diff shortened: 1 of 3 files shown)


royipressburger:
Hi cool plugin!

Does the PR expose the `onprogress` method?

glecaros:
hey, totally missed that, will update tonight

ihadeed:
Hey @glecaros 
Just looking at this now.

How about we implement the new instance wrapper that we have for Ionic Native?

So instead of the `ft` property you have, add a `_objectInstance` property. Then remove any functionality from the instance methods and just add `@CordovaInstance()` decorator to them.

I think this would be a good example to follow: https://github.com/driftyco/ionic-native/blob/master/src/plugins/media.ts

Let me know if you have any questions of suggestions.

Thanks for your efforts.

glecaros:
ok, I'll look into that

ihadeed:
-- For reference, this PR will close #70 if merged.

glecaros:
I've updated to use the CordovaInstance (also rebased to latest master)
I had to change the signature of the wrapInstance method since the parameters didn't match how it was being called.

djangojack:
@glecaros @ihadeed is this rdy to merge?

ihadeed:
Thanks @glecaros

glecaros:
Cool, thanks!

PR #153, opened 2016-05-06
Add typings to hotspot.ts

dudeofawesome:
Adding typings to hotspot.ts, as well as some JSDoc definitions


ihadeed:
Thanks @dudeofawesome for the PR. Great work!

Can you put all the interfaces in the hotspot.ts file instead of separate files?

I understand it's done that way for organizational purposes, but as of now we are just keeping everything in one file until we decide on a specific system/layout.

Thanks again

dudeofawesome:
The new [Angular2 style guide](https://angular.io/docs/ts/latest/guide/style-guide.html#!#single-responsibility) specifies that interfaces should be in their own files. Perhaps, for the sake of organization, we could put them in a new folder called `interfaces`?

ihadeed:
Sorry I wasn't aware of that.

A new folder for interfaces would be good.

Thanks again

dudeofawesome:
Alright, sounds great. I'll update it in a few minutes.

Also, I think I noticed a few misnamed methods in the original wrapper, so I'll fix that too.

ihadeed:
Thank you! Your efforts are appreciated.

dudeofawesome:
Ok, I've made the changes, (https://github.com/dudeofawesome/ionic-native/commit/6f50138d9e22292711e4bb3bbba7f2ab9280f55b) but I'm not sure how to update this PR with those new changes.

ihadeed:
If you update the branch of this PR it will update the PR automatically.

dudeofawesome:
Ok, I think it's ready now!

ihadeed:
Ok thanks I will have a look at it when I'm home.

ihadeed:
Looks good to me. Thank you @dudeofawesome

ihadeed:
Hey @dudeofawesome ... I think we should replace the `.interface` suffix with `.type`. See here: https://angular.io/docs/ts/latest/guide/style-guide.html#naming

Any thoughts?

dudeofawesome:
Yeah, it does look like `.type` would be the preferred suffix.

dudeofawesome:
Hey, when can I expect the fixes I added to make it into a release?

ihadeed:
I will release a patch tonight or tomorrow to fix those little issues.

dudeofawesome:
Great! Thank you!

ihadeed:
@dudeofawesome released!

dudeofawesome:
Awesome, thank you!

image

PR #157, opened 2016-05-16
implement wrappers for Insomnia

matiastucci:


ihadeed:
Thanks for the PR @matiastucci !

PR #166, opened 2016-05-19
type return value of globalization's stringToNumber

rapropos:
This allows client code to compile properly when TypeScript's `noImplicitAny` is enabled.


ihadeed:
Maybe it should be `number|string` instead?

Correct me if I'm wrong, but I guess if you pass a second parameter to the function and set `type: 'currency'` you will get a string back, not a number.

rapropos:
Thank you for noticing that. Serves me right to trust function names.

ihadeed:
thanks @rapropos

PR #196, opened 2016-06-08
Add options while creating map

ZiFFeL1992:


ihadeed:
Hey @ZiFFeL1992 
Thanks for the PR

I think there is no need for the `if` statement. You can pass options either way to the `getMap` function. It should be fine because if you look at the original plugin, the developer uses the following line of code: `params = params || {};`

Issue #101, opened 2016-03-31
background-geolocation


4ware:
I would love to migrate our app to ionic2 but this could not be done without this plugin.

keithdmoore:
@4ware Are you talking about this plugin?  https://github.com/transistorsoft/cordova-background-geolocation-lt

4ware:
@keithdmoore It might be wiser to use https://github.com/mauron85/cordova-plugin-background-geolocation since the original plugin moved to some paid "premium" scheme now.

vfdev-5:
I'd like to try write a wrapper in my [fork](https://github.com/vfdev-5/ionic-native)
However, I do not understand how can I test it with an app...

ihadeed:
hey @vfdev-5 

I see you added the interfaces for now. Next step would be to add wrappers for the functions, then add the plugin to index.ts.

Once you're done with that run `npm run build_bundle` to compile the library and have something that you can import in a Cordova project and test.

vfdev-5:
Hi @ihadeed 
I added wrappers for the functions and the plugin to the index.ts as you said. The compilation is OK and I 'tested' it with a demo application - ionic 2 version of [this](https://github.com/mauron85/cordova-plugin-background-geolocation/tree/master/example/SampleApp)
What is the next step, pull-request, ... ?

ihadeed:
@vfdev-5 yes if the functionality works submit a PR and I'll review & merge.

joshgarwood:
Is ionic-native now using this plugin instead of the one from mauron85?  I recently got the latest ionic2 build (beta8), installed the mauron85 plugin, and imported ionic-native, but it looks like the configure function is now a promise instead of an observable:

```
    var backgroundOptions = {
            desiredAccuracy: 10,
            stationaryRadius: 10,
            distanceFilter: 30,
            debug: false
        };

        // This seems to work
        BackgroundGeolocation.configure(backgroundOptions).then((location) => {
            this.notifyLocation(location);
            BackgroundGeolocation.finish();
        });


        // this does not
        BackgroundGeolocation.configure(
            location => {
                //this.notifyLocation(location);
                BackgroundGeolocation.finish();
            }, 
            err => {
                console.log(err);
            }, 
            backgroundOptions);
```

The [documentation suggests](http://ionicframework.com/docs/v2/native/backgroundgeolocation/) that the second implementation should work.  So I'm just trying to figure out the proper way to do this.  Any help would be much appreciated.

Thanks!

[...]

PR #199, opened 2016-06-09
Fix documentation problem in BackgroundGeolocation plugin

vfdev-5:
As indicated in the issue [#101](https://github.com/driftyco/ionic-native/issues/101), there is a problem with the documentation of BackgroundGeolocation plugin. This PR fixes it.


ihadeed:
Thanks!

PR #205, opened 2016-06-10
Fix #186

### LINKED ISSUE #186: Media Plugin success, error, status handling
Noticed in the code that there is a TODO for handling these.  Is there any progress on this or any hints as to how it could be handled?

I need these call backs when using the media plugin to know the current loading/playback status and whether it was successful or not.

ghenry22:
Update docs for Media Plugin with more comprehensive examples and some platform specific quirks.

Implemented success & error callbacks as promise (thoroughly tested with streams and local files on both iOS and Android)

Implemented status callback as observable (currently not working).


ihadeed:
Thanks @ghenry22 :grin:

PR #213, opened 2016-06-11
chore(ionic-native): remove extras

ihadeed:
This is to remove all the extra files & folders that aren't being used. Even if we were to use them, their content would most likely change. So I think it makes sense to get rid of them for now.

Deleted the following:
- /bundle.js
- /demo/
- /ng1demo/
- /test/

Merge if you are ok with this @mlynch @tlancina


mlynch:
:+1: Good call

PR #218, opened 2016-06-13
SocialSharing: Add recommended `shareWithOptions` method

FdezRomero:
Starting from version 5.1.0, the [SocialSharing plugin docs](https://github.com/EddyVerbruggen/SocialSharing-PhoneGap-Plugin#using-the-share-sheet) recommend to use the new method `shareWithOptions`, that takes an `options` object instead of the 4 arguments and returns a `result` object on success and an error message on failure. The result may include the app selected by the user for analytics purposes (currently iOS only).

This is much more convenient for Ionic developers as the current `share` method doesn't return anything.

This PR implements the `shareWithOptions` method with types and returns a promise with either a result or an error.

PR #220, opened 2016-06-13
SocialSharing: Add recommended `shareWithOptions` method

FdezRomero:
Starting from version 5.1.0, the SocialSharing plugin docs recommend to use the new method shareWithOptions, that takes an options object instead of the 4 arguments and returns a result object on success and an error message on failure. The result may include the app selected by the user for analytics purposes (currently iOS only).

This is much more convenient for Ionic developers as the current share method doesn't return anything.

This PR implements the shareWithOptions method with types and returns a promise with either a result or an error.

(Original PR is #218)


ihadeed:
Thank you again

PR #225, opened 2016-06-14
feat(printer): add printer plugin

zakton5:
Added wrapper for the cordova-plugin-printer.

Here is the git repo: https://github.com/katzer/cordova-plugin-printer.git

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -124,7 +124,6 @@ export {
   Insomnia,
   Keyboard,
   Network,
-  Printer,
   Screenshot,
   SocialSharing,
   Splashscreen,
--- a/src/index.ts
+++ b/src/index.ts
@@ -52,6 +52,7 @@ import {LaunchNavigator} from './plugins/launchnavigator';
 import {LocalNotifications} from './plugins/localnotifications';
 import {MediaPlugin} from './plugins/media';
 import {Network, Connection} from './plugins/network';
+import {Printer} from './plugins/printer';
 import {Push} from './plugins/push';
 import {SafariViewController} from './plugins/safari-view-controller';
 import {Screenshot} from './plugins/screenshot';
@@ -85,6 +86,7 @@ export * from './plugins/inappbrowser';
 export * from './plugins/launchnavigator';
 export * from './plugins/localnotifications';
 export * from './plugins/media';
+export * from './plugins/printer';
 export * from './plugins/push';
 export * from './plugins/safari-view-controller';
 export * from './plugins/sms';
@@ -122,6 +124,7 @@ export {
   Insomnia,
   Keyboard,
   Network,
+  Printer,
   Screenshot,
   SocialSharing,
   Splashscreen,
@@ -183,6 +186,7 @@ window['IonicNative'] = {
   LocalNotifications: LocalNotifications,
   MediaPlugin: MediaPlugin,
   Network: Network,
+  Printer: Printer,
   Push: Push,
   SafariViewController: SafariViewController,
   Screenshot: Screenshot,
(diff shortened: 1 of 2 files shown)


src/index.ts
```diff
@@ -122,6 +124,7 @@ export {
   Insomnia,
   Keyboard,
   Network,
+  Printer,
```
ihadeed: Remove this line as it's not needed. `export *` takes care of this

src/plugins/printer.ts
```diff
@@ -0,0 +1,76 @@
...
+    platforms: ['Android', 'iOS']
+})
+export class Printer {
+
+    /**
+     * Checks whether to device is capable of printing.
+     */
+    @Cordova()
+    static isAvailable(): Promise<boolean> {
+        return new Promise((resolve, reject) => {
```
ihadeed: No need to return a promise manually. Also the original plugin doesn't have Promise support.

The `@Cordova` decorator overrides your function anyways. So this will still work. But you should replace the return here and below with just `return; `. 

The original plugin usage is: 
`isAvailable(callback)`

src/plugins/printer.ts
```diff
@@ -0,0 +1,76 @@
...
+    }
+
+    /**
+     * Sends content to the printer.
+     * @param {content} The content to print. Can be a URL or an HTML string. If a HTML DOM Object is provided, its innerHtml property value will be used.
+     * @param {options} The options to pass to the printer
+     */
+    @Cordova()
+    static print(content: string | HTMLElement, options?: PrintOptions): Promise<any> {
+        return new Promise((resolve, reject) => {
```
ihadeed: Replace the whole return statement with just `return; ` . See comment above.

zakton5:
In summary, both functions should just contain `return;` and have parameters for callbacks? Should the the return signatures for both functions remain as Promises?

ihadeed:
**TL;DR:** No extra parameters are needed, just replace the inner functionality with `return; `

---
**Explaination:**
The `@Cordova()` decorator takes care of the whole process. The default behavior for the decorator is to return a promise. The resolve and reject will be added as arguments to your function, all the way at the end, after every argument you have there.

So for example, let's assume we have a plugin that has the following method:
```javascript
plugins.myPlugin.myMethod = function(optionA, optionB, successCallback, errorCallback) {
  // do stuff here
  try {
    successCallback('return useful info');
  } catch (e) {
    errorCallback('return error');
  }
};
```
To wrap this up with `ionic-native`, we would write the following, and `ionic-native` will do the magic:
```typescript
@Plugin({
  plugin: 'cordova-my-plugin',
  pluginRef: 'plugins.myPlugin',
  repo: 'https://my.github/myuser/myrepo'
})
export class MyPlugin {
  
  /**
  * Some useful docs here!
  */
  @Cordova()
  static myMethod(nameForOptionA: string, nameForOptionB: boolean): Promise<any> {return; }

}
```
Now if we call that method from our app. It will do the following:
1. Check that `plugins.myPlugin` exists, if it doesn't, that means the plugin isn't installed.
2. Creates a Promise (since it's the default behavior and we didn't pass config)
3. Adds the `resolve` and `reject` variables from the promise, to the original plugin's function
4. It calls the original method, and then the resolve or the reject will be triggered and it will transmit the data back to your app

You can configure the decorator to behave differently in case the original plugin is written in a different format.

So basically, when writing a wrapper, you don't need to add any functionality. That is because the decorators we use overwrites your method functionality. There are special cases, like defining a wrapper for a property, or defining an instance based plugin.

Thank you for contributing :grin:

zakton5:
Great explanation! Thank you for all of that info, it is really appreciated. I'll push a new commit in a couple of minutes

zakton5:
There we go. Please let me know if I still missed something

ihadeed:
Looks good now. Thanks @zakton5 !

PR #228, opened 2016-06-15
Changed SafariViewController show return type

Bouzmine:
Changed it to `Promise<any>` as in the example. Was `void`.


ihadeed:
Thanks @Bouzmine !

Issue #245, opened 2016-06-26
Add InAppBrowserRef and InAppBrowserEvents to exports.


SunburnedGoose:
I cannot import `InAppBrowserRef` and `InAppBrowserEvents` into my project because their types are not exported by ionic-native.  I need the `InAppBrowserRef` to get intellisense for the methods on the reference.

    import { InAppBrowser } from 'ionic-native';

    @Injectable()
    export class AuthServices {
        private inAppBrowserRef = undefined;

What I had hoped I could do is...

    import { InAppBrowser, InAppBrowserRef } from 'ionic-native';

    @Injectable()
    export class AuthServices {
        private inAppBrowserRef: InAppBrowserRef = undefined;

ihadeed:
Hey there,

Those interfaces are being exported [here](https://github.com/driftyco/ionic-native/blob/master/src/index.ts#L85).

Please check that you are using the latest version of:
- `ionic-native` (currently at 1.3.0). Run `npm i --save ionic-native@latest`
- IDE and/or TypeScript plugin in IDE

Here's a screenshot of my experiment: 
![image](https://cloud.githubusercontent.com/assets/13794420/16363041/6a7a88c4-3b8c-11e6-9020-1f86c95d5fbe.png)

SunburnedGoose:
Thanks, sorry to trouble you.  I wasn't using 1.3.0.

PR #246, opened 2016-06-26
Adds InAppBrowserRef and InAppBrowserEvent to the types that can be imported into an app.

SunburnedGoose:
This is a PR for issue #245.  I imported two interfaces from InAppBrowser into the file, and then export them.


src/index.ts
```diff
@@ -81,7 +81,6 @@ export * from './plugins/deviceorientation';
 export * from './plugins/googlemaps';
 export * from './plugins/httpd';
 export * from './plugins/imagepicker';
-export * from './plugins/inappbrowser';
```
ihadeed: This line exports everything in `inappbrowser.ts`
I highly doubt that replacing this with individual imports & exports would make a difference.

ihadeed:
See my reply to the issue #245

PR #287, opened 2016-07-13
Update spinnerdialog.ts to fix wrong documentation

nickkenens:
Just implemented the SpinnerDialog into my app to use with Ionic Deploy.

The cancelCallback still happens when set to "false", when setting it to "true, the cancelCallback doesn't get called and therefore the SpinnerDialog doesn't hide on Android, exactly as expected.


ihadeed:
Thank you!

Issue #289, opened 2016-07-13
Improvement to tslint.json


ramonornela:
Change [tslint.json](https://github.com/driftyco/ionic-native/blob/master/tslint.json) to extends and use [tslint-ionic-rules](https://github.com/driftyco/tslint-ionic-rules)

PR #277, opened 2016-07-07
Reformat to match guidelines (WIP)

ZiFFeL1992:

PR #276, opened 2016-07-07
chore(tslint-rules): Use ionic-tslint-rules

ZiFFeL1992:
closes #289

PR #290, opened 2016-07-13
Fix issue #289

ramonornela:


ihadeed:
We already have a PR for this. Thanks though.

ihadeed:
We have this in #276 , waiting on #277 to be fixed then we will merge both PRs.

PR #305, opened 2016-07-17
feat(inappbrowser): implement instance based wrapper

### LINKED ISSUE #247: Rewrite InAppBrowser as instance based plugin
This issue is just here as a reference/reminder.

---

InAppBrowser plugin should be rewritten to implement the `@CordovaInstance` wrapper.

This will provide better coding experience, and allow easier change detection for Angular2 apps.

ihadeed:
This wrapper makes the inappbrowser instance based instead of being static.

The `.open()` method that we currently have returns `InAppBrowserRef` which had some functions that take callbacks as parameters. And since one of the major goals of ionic-native is to convert all callbacks to promises and observables, this wrapper seem to make more sense.

I also added a fallback that will use `window.open()` in a scenario where `cordova` or `cordova.InAppBrowser` is not available. 

Closes #247


src/plugins/inappbrowser.ts
```diff
@@ -12,76 +13,72 @@ export interface InAppBrowserEvent extends Event {
...
+   * A method that allows you to listen to events happening in the browser.
+   * @param event Event name
+   * @returns {Observable<any>} Returns back an observable that will listen to the event on subscribe, and will stop listening to the event on unsubscribe.
    */
-  @Cordova({
-    sync: true
-  })
-  static open(url: string, target?: string, options?: string): InAppBrowserRef { return; }
-
+  on(event: string): Observable<InAppBrowserEvent> {
```
mlynch: What's the purpose of this function? Is this generic and something we could abstract for other plugins?

src/plugins/inappbrowser.ts
```diff
@@ -12,76 +13,72 @@ export interface InAppBrowserEvent extends Event {
   message: string;
 }
```
mlynch: Could you add a quick comment on very basic usage of this? The instance method stuff is a bit new for people and we haven't talked a ton about it.

mlynch:
Looks like a nice improvement. Added a few comments on a few lines. Basically, it needs a simple usage comment and then I was confused by `on()` feeling like a generic thing that we might want to export as an abstraction for other plugins

ihadeed:
The `on` function wraps this function here: https://github.com/apache/cordova-plugin-inappbrowser#inappbrowseraddeventlistener

PR #320, opened 2016-07-17
fix(contacts): missing name property, types

Manduro:
The `@InstanceProperty` for `name` was missing. This made it impossible to set the name of a contact properly, rendering this plugin pretty much useless for creating contacts.

There were also some type issues:
1. `nickname` should be type `string`
2. `phoneNumbers` should be type `IContactField[]`
3. All types should use interfaces, as the objects themselves have an `_objectInstance` property.
4. `Contact` should implement `IContactProperties`

These bugs came in at v1.3.0, where this plugin was refactored, although the changelog does not mention this.


ihadeed:
thanks @Manduro

PR #337, opened 2016-07-20
feat(full-screen-image): add FullScreenImage plugin

ihadeed:


ihadeed:
Tested on Android only, works great.

Repo says it's tested on iOS.

ihadeed:
### I think this is better: https://github.com/sarriaroman/photoviewer

PR #341, opened 2016-07-20
Add hasPermission example to Push docs

ktross:
This works a bit differently than [phonegap-plugin-push](https://github.com/phonegap/phonegap-plugin-push), so I would like to add a very simple example to the Push docs.


ihadeed:
We're not keeping usage in the JSDoc for each function. All usage info should be in the `@usage` or `@advance` sections at the top (in the docs for the `@Plugin` decorator.

PR #346, opened 2016-07-21
Fixes bug in readAsText() and adds readAsDataURL() in File plugin

amuramoto:


ihadeed:
We have a new file plugin in the PR section.

amuramoto:
@ihadeed Would it not be advisable to merge so that the current version works properly while the new plugin is being finished?

ihadeed:
Sure, you're right.

PR #347, opened 2016-07-22
Add Secure Storage

begriffin:
Plugin for https://github.com/Crypho/cordova-plugin-secure-storage

### DIFF
--- a/src/plugins/securestorage.ts
+++ b/src/plugins/securestorage.ts
@@ -29,7 +29,7 @@ declare var cordova: any;
  * 
  * secureStorage.set('myitem', 'myvalue')
  *  .then(
- *    data => console.log( data),
+ *    data => console.log(data),
  *    error => console.log(error)
  * );
  * 
@@ -48,9 +48,7 @@ declare var cordova: any;
 })
 export class SecureStorage {
 
-  // Properties
   private _objectInstance: any;
-  init: Promise<any>;
 
   constructor()  {}
 
--- a/src/plugins/securestorage.ts
+++ b/src/plugins/securestorage.ts
@@ -1,0 +1,87 @@
+import { CordovaInstance, Plugin } from './plugin';
+
+declare var cordova: any;
+
+/**
+ * @name Secure Storage
+ * @description
+ * This plugin gets, sets and removes key,value pairs from a device's secure storage.
+ *
+ * Requires Cordova plugin: `cordova-plugin-secure-storage`. For more info, please see the [Cordova Secure Storage docs](https://github.com/Crypho/cordova-plugin-secure-storage).
+ *
+ * @usage
+ * 
+ * ```typescript
+ * import { SecureStorage } from 'ionic-native';
+ *
+ * let secureStorage = SecureStorage.init('my_app_name');
+ * 
+ * secureStorage.get('myitem')
+ *  .then(
+ *    data => console.log(data),
+ *    error => console.log(error)
+ * );
+ * 
+ * secureStorage.set('myitem', 'myvalue')
+ *  .then(
+ *    data => console.log( data),
+ *    error => console.log(error)
+ * );
+ * 
+ * secureStorage.remove('myitem')
+ * .then(
+ *    data => console.log(data),
+ *    error => console.log(error)
+ * );
+ * ```
+ */
+@Plugin({
+  plugin: 'cordova-plugin-secure-storage',
+  pluginRef: 'plugins.securestorage',
+  repo: 'https://github.com/Crypho/cordova-plugin-secure-storage',
+  platforms: ['Android', 'iOS', 'Windows Phone', 'Browser']
+})
+export class SecureStorage {
+
+  // Properties
+  private _objectInstance: any;
+  init: Promise<any>;
+
+  /**
+   * Creates a namespaced storage. 
+   * @param storage {string}
+   */
+  constructor(storage: string) {
+    let res, rej
+    this.init = new Promise<any>((resolve, reject) => { res = resolve; rej = reject; });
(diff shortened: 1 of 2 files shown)


src/plugins/securestorage.ts
```diff
@@ -0,0 +1,87 @@
...
+  // Properties
+  private _objectInstance: any;
+  init: Promise<any>;
+
+  /**
+   * Creates a namespaced storage. 
+   * @param storage {string}
+   */
+  constructor(storage: string) {
+    let res, rej
```
marcinwadon: You forgot `;`

src/plugins/securestorage.ts
```diff
@@ -0,0 +1,87 @@
...
+ * This plugin gets, sets and removes key,value pairs from a device's secure storage.
+ *
+ * Requires Cordova plugin: `cordova-plugin-secure-storage`. For more info, please see the [Cordova Secure Storage docs](https://github.com/Crypho/cordova-plugin-secure-storage).
+ *
+ * @usage
+ * 
+ * ```typescript
+ * import { SecureStorage } from 'ionic-native';
+ *
+ * let secureStorage = SecureStorage.init('my_app_name');
```
ihadeed: Shouldn't we do something like this:
```typescript
let secureStorage = new secureStorage();
secureStorage.init().then(()=>{ // it's ready }).catch(()=>{ // something wrong });
```

ihadeed:
Read the comments.

Also I think we should follow what Max did with the SQLite plugin here: https://github.com/driftyco/ionic-native/commit/f1c8ce35e85d9ddebf7288e9c1abb5e70811d23d

This way we can keep the ionic-native modifications consistent.

The usage would be something like:
```typescript
let secureStorage: SecureStorage = new SecureStorage();
secureStorage.create(<options>).then(<onSuccess>).catch(<onError>);
```

src/plugins/securestorage.ts
```diff
@@ -47,14 +52,16 @@ export class SecureStorage {
   private _objectInstance: any;
   init: Promise<any>;
```
ihadeed: Forgot to remove this

src/plugins/securestorage.ts
```diff
@@ -47,14 +52,16 @@ export class SecureStorage {
...
 
+  constructor()  {}
```
ihadeed: minor: double space should be space

src/plugins/securestorage.ts
```diff
@@ -47,14 +52,16 @@ export class SecureStorage {
...
+
   /**
    * Creates a namespaced storage. 
-   * @param storage {string}
+   * @param store {string}
```
ihadeed: I think it's better to set this param name to `name` instead of `store`, don't you think?

src/plugins/securestorage.ts
```diff
@@ -0,0 +1,94 @@
...
+  plugin: 'cordova-plugin-secure-storage',
+  pluginRef: 'plugins.securestorage',
+  repo: 'https://github.com/Crypho/cordova-plugin-secure-storage',
+  platforms: ['Android', 'iOS', 'Windows Phone', 'Browser']
+})
+export class SecureStorage {
+
+  // Properties
+  private _objectInstance: any;
+  init: Promise<any>;
```
ihadeed: this is unecessary now

src/plugins/securestorage.ts
```diff
@@ -0,0 +1,94 @@
...
+ * 
+ * secureStorage.get('myitem')
+ *  .then(
+ *    data => console.log(data),
+ *    error => console.log(error)
+ * );
+ * 
+ * secureStorage.set('myitem', 'myvalue')
+ *  .then(
+ *    data => console.log( data),
```
marcinwadon: Please remove this space after the first parenthesis

src/plugins/securestorage.ts
```diff
@@ -0,0 +1,94 @@
...
+@Plugin({
+  plugin: 'cordova-plugin-secure-storage',
+  pluginRef: 'plugins.securestorage',
+  repo: 'https://github.com/Crypho/cordova-plugin-secure-storage',
+  platforms: ['Android', 'iOS', 'Windows Phone', 'Browser']
+})
+export class SecureStorage {
+
+  // Properties
+  private _objectInstance: any;
```
marcinwadon: There is no need to write `_` prefix because typescript knows that this is a private member. What do you think?

src/plugins/securestorage.ts
```diff
@@ -0,0 +1,94 @@
...
+  init: Promise<any>;
+
+  constructor()  {}
+
+  /**
+   * Creates a namespaced storage. 
+   * @param store {string}
+   */
+  create(store: string): Promise<any> {
+    return new Promise((res, rej) => {
```
marcinwadon: What types of this parameters are? Shouldn't be declare as eg. `Function`?

ihadeed: This is how I designed the `CordovaInstance` decorator to work. To change this we would have to modify the other plugins. We can keep this for now and maybe modify all the plugins later.

src/plugins/securestorage.ts
ihadeed: @marcinwadon This is how I designed the `CordovaInstance` decorator to work. To change this we would have to modify the other plugins. We can keep this for now and maybe modify all the plugins later.

ihadeed:
Looks good @begriffin 
Thank you for the PR!

ihadeed:
I'm thinking the `create` function should be refactored to something like `openStorage`. Since that function would open the database if the name exists. (similar to SQLite behaviour).

What do you guys think?

PR #349, opened 2016-07-22
example added for Transfer plugin

sameera207:
Hi,

Had a bit of difficulty on finding a complete example on using the [Transfer plugin](http://ionicframework.com/docs/v2/native/transfer/).  

I finally got it working and thought it would be nice to have a full example, so it might help someone in the future. 

I believe adding an example as a comment, will show that in the [documentation page](http://ionicframework.com/docs/v2/native/transfer/). If not please let me know where to do the changes :). 

cheers
Sam

### DIFF
--- a/src/plugins/filetransfer.ts
+++ b/src/plugins/filetransfer.ts
@@ -127,6 +127,27 @@ export interface FileTransferError {
  *
  * // Abort active transfer:
  * fileTransfer.abort();
+ *
+ * E.g
+ *
+ * upload(){
+ *   const fileTransfer = new Transfer();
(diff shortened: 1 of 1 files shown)


ihadeed:
Thanks @sameera207

Sampath-Lokuge:
Good job @sameera207 

But that example shows this error no? Do you know why? [**Ionic native Transfer plugin's `file.dataDirectory` shows error**](https://stackoverflow.com/questions/44428434/ionic-native-transfer-plugins-file-datadirectory-shows-error)

PR #351, opened 2016-07-25
Update statusbar.ts

ankurvr:
Typo change in Usage example

[Docs Page](http://ionicframework.com/docs/v2/native/status-bar/)  Here in Usage section 
**StatuBar.overlaysWebView(true);** 
to 
**StatusBar.overlaysWebView(true);**


ihadeed:
thanks!

PR #352, opened 2016-07-25
OneSignal improvements

Inoverse:
Two OneSignal changes:
* Changed init function to support notificationOpenedCallback as an optional parameter
* Removed platform specification


ihadeed:
Hey there,

Why did you change the Promise to a Callback?

Is it because that callback gets triggered multiple times? If that's the case we can convert the return type to an Observable instead.

Inoverse:
Hi,

yes it's not trigged during the application run. It's triggered when a notification was received, which can be occures multiple times. Observable could be a solution.

ihadeed:
`@Corova({
  observable: true
})
static methodName(): Observable<any> {return; }
`

Inoverse:
Changed it to an observable.

ihadeed:
thanks

PR #356, opened 2016-07-27
feat(estimote-beacons): Add Estimote beacons plugin

fabLouis:
Plugin repository: https://github.com/evothings/phonegap-estimotebeacons

TODO: the sub-plugin for the Estimote nearables

### DIFF
--- a/src/plugins/estimote-beacons.ts
+++ b/src/plugins/estimote-beacons.ts
@@ -17,59 +17,59 @@ import { Observable } from 'rxjs/Observable';
 export class EstimoteBeacons {
 
   /** Proximity value */
-  public static ProximityUnknown = 0;
+  static ProximityUnknown = 0;
 
   /** Proximity value */
-  public static ProximityImmediate = 1;
+  static ProximityImmediate = 1;
 
   /** Proximity value */
-  public static ProximityNear = 2;
+  static ProximityNear = 2;
 
   /** Proximity value */
-  public static ProximityFar = 3;
+  static ProximityFar = 3;
 
   /** Beacon colour */
-  public static BeaconColorUnknown = 0;
+  static BeaconColorUnknown = 0;
 
   /** Beacon colour */
-  public static BeaconColorMintCocktail = 1;
+  static BeaconColorMintCocktail = 1;
 
   /** Beacon colour */
-  public static BeaconColorIcyMarshmallow = 2;
+  static BeaconColorIcyMarshmallow = 2;
 
   /** Beacon colour */
-  public static BeaconColorBlueberryPie = 3;
+  static BeaconColorBlueberryPie = 3;
   /**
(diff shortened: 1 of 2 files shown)


src/plugins/estimote-beacons.ts
```diff
@@ -0,0 +1,525 @@
...
+  /**
+   * Start ranging secure beacons. Available on iOS.
+   * This function has the same parameters/behaviour as
+   * {@link EstimoteBeacons.startRangingBeaconsInRegion}.
+   * To use secure beacons set the App ID and App Token using
+   * {@link EstimoteBeacons.setupAppIDAndAppToken}.
+   */
+  @Cordova({
+    observable: true,
+    clearFunction: 'stopRangingSecureBeaconsInRegion'
```
ihadeed: this probably needs : `clearWithArgs: true` as well.

ihadeed:
Hey this looks great

I just think the observables all need to have `clearWithArgs: true` since the plugin needs it, see https://github.com/evothings/phonegap-estimotebeacons/blob/master/plugin/src/js/EstimoteBeacons.js#L726

fabLouis:
Ok thanks @ihadeed but before do you have more information about this argument? It will be interesting to complete the [Ionic Native Developer Guide](https://github.com/driftyco/ionic-native/blob/master/DEVELOPER.md) page?

ihadeed:
It would look like this:
```typescript
 @Cordova({
    observable: true,
    clearFunction: 'stopSomeMethod',
    clearWithArgs: true
  })
  static startSomeMethod(someArg): Observable<any> { return; }
```
It applies the same arguments that were supplied to the `start` function when cancelling the watch. 

See: https://github.com/driftyco/ionic-native/blob/master/src/plugins/plugin.ts#L139-L144

I'll try to add docs to the developer guide soon.

src/plugins/estimote-beacons.ts
```diff
@@ -0,0 +1,529 @@
...
+   * @property {string} identifier Region identifier
+   * (id set by the application, not related actual beacons).
+   * @property {string} uuid The UUID of the region.
+   * @property {number} major The UUID major value of the region.
+   * @property {number} major The UUID minor value of the region.
+   * @property {string} state One of
+   * {@link estimote.beacons.RegionStateInside},
+   * {@link estimote.beacons.RegionStateOutside},
+   * {@link estimote.beacons.RegionStateUnknown}.
+   */
```
ihadeed: What are those extra docs for?

ihadeed:
Hey I just had another look

1. The default for TypeScript is `public` so there is no need to state that before the property names
2. Saw some extra docs in the file, are they supposed to explain something? If so, it's better to move them somewhere more appropriate.

ihadeed:
thanks @fabLouis

PR #366, opened 2016-07-29
Added Screen Orientation Plugin #342

### LINKED ISSUE #342: screen orientation plugin
it would be great if we could use https://github.com/apache/cordova-plugin-screen-orientation through ionic-native

Inoverse:
I've implemented the screen orientation plugin which closes #342


ihadeed:
thanks @Inoverse

PR #370, opened 2016-07-30
Fix quotes on TouchID.isAvailable error comment

johncblandii:
You can see the documentation code coloring is off due to the quotation marks. I went with `"` instead of escaping.

screenshot 2016-07-30 14 07 09


ihadeed:
thanks @johncblandii

johncblandii:
No prob, @ihadeed.

Issue #224, opened 2016-06-14
Implement remaining functions (Diagnostic plugin)


ZiFFeL1992:
There are some remaining functions to be wrapped.

Repo: https://github.com/dpa99c/cordova-diagnostic-plugin

PR #373, opened 2016-07-31
fix(diagnostic): corrects isEnabled functions to isAvailable

amuramoto:
* Location, camera, wifi, and bluetooth 'isEnabled' functions are incorrect. Corrected all to 'is Available'.

### DIFF
--- a/src/plugins/diagnostic.ts
+++ b/src/plugins/diagnostic.ts
@@ -11,28 +11,28 @@ export class Diagnostic {
    * Checks if app is able to access device location.
    */
   @Cordova()
-  static isLocationEnabled(): Promise<any> { return; }
+  static isLocationAvailable(): Promise<any> { return; }
 
   /**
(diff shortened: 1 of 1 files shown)


amuramoto:
Fixes https://github.com/driftyco/ionic-native/issues/365

ihadeed:
Hey there,

Thanks for the PR.

The methods shouldn't be replaced. The plugin actually exports both `isLocationAvailable` and `isLocationEnabled`. 

The current implementation of the plugin is just missing many of the functions listed here: https://www.npmjs.com/package/cordova.plugins.diagnostic

Related:  https://github.com/driftyco/ionic-native/issues/224

ihadeed:
Just checked the docs again, you're right these should be `Available` instead of `Enabled`. But we will still be missing the `Enabled` functions for now. 

Thanks again!

PR #376, opened 2016-07-31
Update inappbrowser.ts

nazrdogan:


ihadeed:
Thanks for the PR.

I'll merge this but we are switching to a new implementation for the InAppBrowser soon.

Issue #367, opened 2016-07-29
Readme: ionic-native vs ngCordova and $scope.$apply


mirkonasato:
The readme currently says that for [Angular 1](https://github.com/driftyco/ionic-native#angular-1):

> Ionic Native works as a stand-in for ngCordova. In many cases, the usage is identical, but we import ionic.native instead of ngCordova as our module.

That doesn't seem very accurate: looks like with ionic-native you always need to wrap updates in `$scope.$apply`. That's usually not required with ngCordova because it uses the $q service.

ihadeed:
Should be fixed in next release.

mirkonasato:
Tried 1.3.10 but I still need `$scope.$apply` with e.g. `$cordovaCamera.getPicture()`.

Maybe it's because #378 says it's a "partial fix"?

Issue #348, opened 2016-07-22
Observable Usage with Angular 1.x?


MattHoneycutt:
Do observables (such as the return of Geolocation.watchPosition()) work with Angular 1.x's digest loop?  Or do I manually need to fire a $scope.$digest() in my subscribe callback?

I've got this subscription:

```jscript
            subscription = $cordovaGeolocation.watchPosition().subscribe(data => {
                vm.latitude = data.coords.latitude;
                vm.longitude = data.coords.longitude;
                console.log(data, vm);
            });
```

I can see the data being assigned, but it never refreshes the view, because a digest loop is never started.

MattHoneycutt:
Actually, it looks like this question would apply to promises as well.  Do I need to manually call $scope.$digest or $scope.$apply?  Should something else be calling them?  Or is there another elegant solution to this problem for Angular 1.x that I'm overlooking?

ihadeed:
For promises, you can just use them directly without having to modify anything. 

Before wrapping any method call, Ionic Native attempts to find ES6 Promise support, if that's not found, it will attempt to find `window.angular` (if found, will use $q), and if that's not found either it will throw a `console.warn` to let you know that you need to add a polyfill for promise support.

As for observables, I haven't done any experiments with that yet but I guess that it should work out of the box since the files `ionic.native.js` and `ionic.native.min.js` are produced by a browserify command that should pull all the required code from `rxjs` package.

I will close this for now, but do reopen or comment if you have further questions or difficulties.

MattHoneycutt:
That doesn't seem to be the case.  I'm using ionic-native 1.3.9 with Angular 1.5.7.  Here's my code using the $cordovaGeolocation plugin:

```jscript
                $cordovaGeolocation.getCurrentPosition().then(data => {
                    console.log('Initial get completed.', data);
                    vm.latitude = data.coords.latitude.toFixed(4);
                    vm.longitude = data.coords.longitude.toFixed(4);
                });
```
My console.log is executed, but my view never updates, because a digest is never triggered.  

If I wrap the call in $scope.$apply, the view updates as expected:

```jscript
            $cordovaGeolocation.getCurrentPosition().then(data => {
                $scope.$apply(() => {
                    console.log('Initial get completed.', data);
                    vm.latitude = data.coords.latitude.toFixed(4);
                    vm.longitude = data.coords.longitude.toFixed(4);
                });
            });
```

Am I doing something wrong?

mlynch:
[...]

PR #378, opened 2016-07-31
fix(ng1): use $q promises instead of the native Promise

mattlewis92:
This is a partial fix for #348 and #367 (ES6 promise's are currently given priority over ng1's $q so a scope.$apply is required for all promise callbacks as $q is never actually used).

The other half involves using an observable to trigger a digest which should be possible by using the rxjs [scope scheduler](https://github.com/Reactive-Extensions/rx.angular.js/tree/master/docs#rxscopescheduler).


ihadeed:
thanks!

PR #384, opened 2016-08-02
Update nativestorage.ts

GillesC:
Small adjustments and additions to the doc.

### DIFF
--- a/src/plugins/nativestorage.ts
+++ b/src/plugins/nativestorage.ts
@@ -2,8 +2,8 @@ import { Cordova, Plugin } from './plugin';
 
 
 /**
- * @name Native Storage
- * @description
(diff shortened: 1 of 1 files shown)


src/plugins/nativestorage.ts
```diff
@@ -2,8 +2,8 @@ import { Cordova, Plugin } from './plugin';
 
 
 /**
- * @name Native Storage
- * @description
+ * @name NativeStorage
```
ihadeed: This is the name that appears in the docs. We want to keep the spacing in between the words.

GillesC: I'm the developer of the NativeStorage plugin, and I intended to use it as a whole to refer to LocalStorage.

PR #386, opened 2016-08-03
Removed unnecessary ionic.native.min.js

walkingriver:
When installing the bower package created from this, a couple of warnings are printed:

 - bower ionic-native#* invalid-meta The "main" field cannot contain minified files
 - bower ionic-native#* invalid-meta The "main" field has to contain only 1 file per filetype; found multiple .js files: ["ionic.native.js","ionic.native.min.js"]


mlynch:
Thanks!

PR #388, opened 2016-08-03
Using option otherPromise instead of sync

Ritzlgrmft:
Since the plugin's methods return already a promise, the workaround with the `sync` option was used. This worked well for the method calls. However, when either cordova or the plugin was not available, an error was thrown, instead of rejecting the returned promise.

Therefore a better way is to use the `otherPromise` option, introduced with https://github.com/driftyco/ionic-native/releases/tag/v1.3.8.

### DIFF
--- a/src/plugins/ibeacon.ts
+++ b/src/plugins/ibeacon.ts
@@ -419,7 +419,7 @@ export class IBeacon {
    * @return Returns a promise which is resolved as soon as the
    * native layer acknowledged the request and started to send events.
    */
-  @Cordova({sync: true})
+  @Cordova({otherPromise: true})
   static onDomDelegateReady(): Promise<void> { return; }
 
   /**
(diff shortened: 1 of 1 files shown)


mlynch:
Thanks, will have ibrahim review for the next release

ihadeed:
Looks good @Ritzlgrmft, have you tested it?

Ritzlgrmft:
Yes, it is working.

Additionally, I get a rejected promise, if the plugin is not available - as it should be.

Am 04.08.2016 um 23:56 schrieb Ibrahim Hadeed <notifications@github.com<mailto:notifications@github.com>>:


Looks good @Ritzlgrmft<https://github.com/Ritzlgrmft>, have you tested it?

-

ihadeed:
thanks @Ritzlgrmft

PR #392, opened 2016-08-04
docs(google maps) - Updates to Google Maps docs

tlaverdure:
I would do more, but I'll see what kind of feedback I get first from these changes.


mlynch:
Many thanks

PR #393, opened 2016-08-04
Added comment for iOS.

walkingriver:
It doesn't matter what you send as the duration to vibrate() under iOS, it's always the same.


mlynch:
Thanks!

PR #417, opened 2016-08-11
fix(camera): camera options should be optional. 

### LINKED ISSUE #413: CameraOption should be optional
Right now, CameraOptions are a required parameter in `getPicture()`

But according to the cordova docs, CameraOptions are optional .

https://github.com/apache/cordova-plugin-camera#cameracameraoptions--object

Even looking a the older DT definitions, they list it as optional.

https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/cordova/plugins/Camera.d.ts#L38

kessiler:
Fixes #413

### DIFF
--- a/src/plugins/camera.ts
+++ b/src/plugins/camera.ts
@@ -188,13 +188,13 @@ export class Camera {
 
   /**
    * Take a picture or video, or load one from the library.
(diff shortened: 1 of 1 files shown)


ihadeed:
thanks

PR #422, opened 2016-08-15
feat(): wait for deviceready before performing async functions

### LINKED ISSUE #100: Option to wait for device ready event on "@Cordova"-decorator
Shouldn't all plugin-calls wait for the device-ready-promise to be fulfilled?

Or could we at least have an option to delay plugin-calls until the device is ready?

``` ts
import {Plugin, Cordova} from 'ionic-native';
...
export class Test{
    @Cordova(onDeviceReady:true)
    static startTrackerWithId(id:string): Promise<any> { return }
}
```

ihadeed:
Discuss in #100 

Closes #100

### DIFF
--- a/src/plugins/plugin.ts
+++ b/src/plugins/plugin.ts
@@ -84,8 +84,8 @@ function callCordovaPlugin(pluginObj: any, methodName: string, args: any[], opts
   }
 
   if (resolve) {
-    // if we are working with an observable or a promise, let's wait for onDeviceReady since this is an async call
-    document.addEventListener('ondeviceready', () => {
+    // if we are working with an observable or a promise, let's wait for deviceready since this is an async call
+    document.addEventListener('deviceready', () => {
       get(window, pluginObj.pluginRef)[methodName].apply(pluginInstance, args);
     });
   } else {
--- a/src/plugins/plugin.ts
+++ b/src/plugins/plugin.ts
@@ -83,8 +83,15 @@ function callCordovaPlugin(pluginObj: any, methodName: string, args: any[], opts
     };
   }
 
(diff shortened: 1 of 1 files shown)


ramonornela:
@ihadeed I think that can merge in >= 2.0 the current.
Thanks

cfjedimaster:
Any ETA on this one? Sorry to nag.

mlynch:
I'm not sure how I feel about this yet. If you call a plugin and device isn't ready, what should the response to the plugin be? Should it reject the promise? Should it wait for the result? What are the perf implications of that mechanism?

Those are my concerns right now...

cfjedimaster:
I'd vote for waiting since - in theory - it is a very short wait and always a one time thing. (To be clear, I only chimed in here cuz of the Contacts stuff, no idea about the rest. ;)

ihadeed:
I think this PR is good because:
1. It will not break any apps
2. Waiting for `deviceready`, or waiting in general is an expected behaviour from any asynchronous function
3. It shouldn't slow down plugin calls. According to the Cordova documentation, if you wrap something with `deviceready` listener, and the  event has already fired, the callback will be invoked immediately. 

I was planning to merge this for v2.3.0, but if you have any concerns about this @mlynch then we can wait.

cfjedimaster:
Been a month now - any way we can get this in? If not, can a fix for #737 be done in another PR?

ihadeed:
@cfjedimaster i'm currently testing all the new changes for 2.3.0

Just making sure nothing is going to break, then I will merge them all and publish a new release. 

Btw, #737 is going to be fixed by #635

ihadeed:
Will resubmit for v3 if needed

PR #425, opened 2016-08-15
feat(push): plugin is now instance based

### LINKED ISSUE #250: Push Plugin: Add `clearAllNotifications` method and return promises/observables
Hey there!

I started using the Push plugin and noticed that is missing the newer [clearAllNotifications](https://github.com/phonegap/phonegap-plugin-push/blob/master/docs/API.md#pushclearallnotificationssuccesshandler-errorhandler---ios-android-only) method.

I also noticed that the ionic-native wrapper is using the plugin's success and error callbacks instead of returning a promise/observable as it happens with most of the other plugins, which is confusing considering the docs say plugins return either of these.

Any chance to add this missing method and use promises/observables instead of callbacks? I'm guessing there are people already using this plugin so it should keep supporting the callbacks and return a promise/observable if none is specified.

Thanks!

ihadeed:
### Breaking Change

no more callbacks

might need more documenting

closes #250


src/plugins/push.ts
```diff
@@ -265,9 +188,7 @@ export interface PushOptions {
   windows?: {};
 }
 
-declare var PushNotification: {
-  new (): PushNotification
-};
+export type PushEvents = 'registeration' | 'error' | 'notification';
```
Manduro: Typo here: `registeration` should be `registration`

ihadeed: thanks @Manduro I'll get that fixed

ramonornela:
@ihadeed I think that can merge in >= 2.0 the current.
Thanks

ihadeed:
This is now merged in v3-injectable.

Issue #311, opened 2016-07-17
TODO: Add TTS plugin


ihadeed:
https://github.com/vilic/cordova-plugin-tts

jackvial:
+1

PR #431, opened 2016-08-15
feat(TTS): add tts plugin

cmanou:
add TTS to complete #311

### DIFF
--- a/src/plugins/text-to-speech.ts
+++ b/src/plugins/text-to-speech.ts
@@ -26,10 +26,10 @@ export interface TTSOptions {
  */
 @Plugin({
   plugin: 'cordova-plugin-tts',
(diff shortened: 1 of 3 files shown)


src/plugins/text-to-speech.ts
```diff
@@ -0,0 +1,47 @@
...
+ *
+ * TTS.speak('Hello World')
+ *   .then(() => console.log('Success'))
+ *   .catch((reason: any) => console.log(reason));
+ *
+ * ```
+ */
+@Plugin({
+  plugin: 'cordova-plugin-tts',
+  pluginRef: 'window.TTS',
```
ihadeed: `window.` is not needed

ihadeed:
Thanks @cmanou

PR #436, opened 2016-08-16
Add the reject function at the expected errorIndex position in the args array

Boosten:
We don't want that the reject cb takes the position of an optional argument that has not been defined.
For example the Dialogs.alert method takes an optional 'buttonLabel' string. In case we do not set this value, and thus want to use the default value, the 'reject' callback get spliced into this position due the fact that the splice start index is bigger than the array length.

`Dialogs.alert('mytitle', 'mymessage', 'Okay')`
args will become: 
`["title",  resolve, "message", "My button text", reject]`

`Dialogs.alert("title", "message")`
args will become:
`["title", resolve, "message", reject]` 

The latter will invoke the **cordova-plugins-dialog** alert method with the wrong arguments
--> reject is on the position of the buttonLabel

**cordova-plugins-dialog** alert method signature
`alert: function(message, completeCallback, title, buttonLabel) {..}`

### DIFF
--- a/src/plugins/plugin.ts
+++ b/src/plugins/plugin.ts
@@ -51,7 +51,13 @@ function setIndex(args: any[], opts: any = {}, resolve?: Function, reject?: Func
   } else if (typeof opts.successIndex !== 'undefined' || typeof opts.errorIndex !== 'undefined') {
     // If we've specified a success/error index
     args.splice(opts.successIndex, 0, resolve);
-    args.splice(opts.errorIndex, 0, reject);
+
+    // We don't want that the reject cb gets spliced into the position of an optional argument that has not been defined and thus causing non expected behaviour.
+    if (opts.errorIndex > args.length) {
+      args[opts.errorIndex] = reject; //insert the reject fn at the correct specific index
+    } else {
+      args.splice(opts.errorIndex, 0, reject); //otherwise just splice it into the array
+    }
   } else {
     // Otherwise, let's tack them on to the end of the argument list
     // which is 90% of cases


ihadeed:
This is already taken into consideration when it comes to the dialogs plugin: 
image

However, we can still look into merging this as it might help solve future conflicts.

Boosten:
Hi @ihadeed, 

You mean the specific success/error Index? I'm aware of that, but it's not guaranteed that the "errorIndex" also takes that specific index in the args array. 

I encountered this problem in combination with angular 1 and the ionic-native-bower package, but it also occurs in an Ionic 2 app obviously.

ihadeed:
No I meant the default values that are set to the parameters.

I expected them to be passed to the plugin if the developer hasn't passed any values. But it turned out it doesn't work that way, since the decorators override that functionality. This is what it looks like in ES5:
image

Boosten:
Ah, and eventually it will fallback on the cordova-plugin-dialogs implementation default values:

`alert: function(message, completeCallback, title, buttonLabel) {
        var _title = (typeof title === "string" ? title : "Alert");
        var _buttonLabel = (buttonLabel || "OK");
        exec(completeCallback, null, "Notification", "alert", [message, _title, _buttonLabel]);
    }`

where in this case the fourth argument is not a string but it's defined:
`["title", resolve, "message", reject]`

and will not set the **_buttonLabel** to "OK" and invokes the exec command with a wrong value. 

So your're saying the root cause to this problem is actually something else? I'll try to look into it.

ihadeed:
No, the solution you provided works well. It solves the problem.

I think we need to do the same thing for successIndex.

PR #441, opened 2016-08-16
fix(camera-preview): Changes implementation to match Cordova plugin

amuramoto:
- Changes interfaces to match plugin
- Minor docs updates
- Changes startCamera() function signature to match plugin


ihadeed:
Thanks Alex

We keep a prefix for the interface name to make sure that there are no conflicts in the future (in case another plugin uses the same name for their interfaces).

amuramoto:
No problem. I've prefixed the `Size` and `Rect` interfaces

ihadeed:
thanks @amuramoto

PR #442, opened 2016-08-16
feat(geofence): Adds geofence plugin

### LINKED ISSUE #314: TODO: Geofence plugin
https://github.com/cowbell/cordova-plugin-geofence

amuramoto:
Closes #314 

Supports all Cordova plugin functions except:

-onTransitionReceived()
-onNotificationClicked()

These functions require the base window.geofence object to be modified directly. See https://github.com/cowbell/cordova-plugin-geofence#listening-for-geofence-transitions


ihadeed:
Why is there an edit in the camera preview plugin? is that there by accident?

ihadeed:
Also, we can do something like this for the two methods you mentioned

```typescript
// add this to the top of the document
declare var window: any;

static onTransitionReceived(): Observable<any> {

return new Observable<any>((observer) => {
window.geofence.onTransitionReceived = observer.next.bind(observer);
return () => window.geofence.onTransactionReceived = undefined
});

}
```

amuramoto:
Oops must have cut the branch wrong. New commit reverts Camera Preview plugin for this branch to what's in master.

amuramoto:
@ihadeed I tried your suggestion for onTransitionReceived, but it returns a `ZoneAwarePromise` rather than an observer. Any advice?

ihadeed:
Are you using a Cordova decorator? You shouldn't use one

amuramoto:
Removed the Cordova decorator. Function now returns the Observable, but the observer is never called.

ihadeed:
Can you provide a code sample of how you're calling the plugin?

amuramoto:
Here's the plugin code:

```
static onTransitionReceived(): Observable<any> {

    return new Observable<any>((observer) => {
      window.geofence.onTransitionReceived = observer.next.bind(observer);
      return null;
      // undefined can be replaced with ()=>{} .. whichever works better
    });

  }
```

And here's the code calling it, which should update a data binding in the view, but does not appear to:

```
Geofence.onTransitionReceived().subscribe(()=>this.transited='TRANSITED');
```

ihadeed:
Just fixed them. When we provide custom code for a wrapper we can't use `@Cordova` decorator as it overrides any code in our function. Thank @amuramoto

PR #446, opened 2016-08-17
Additions to documentation and type definitions for File plugin

beders:
Added return types to documentation as well as modified return type definition to match actual types returned.
Used union types for promises where applicable.

Builds and gulp docs ok.


ihadeed:
thanks

PR #451, opened 2016-08-18
fix(ng1): remove the need to trigger digests on promise callbacks

mattlewis92:
Currently using `window.angular.injector(['ng']).get('$q');` will create a new angular injector instance which will create a new instance of $q and hence the $rootScope meaning the digests will trigger on a different cycle than the actual app. This is a hacky fix that gets `$q` via the run block and exposes it globally on the `IonicNative` global.

### DIFF
--- a/src/ng1.ts
+++ b/src/ng1.ts
@@ -7,7 +7,9 @@ declare var window;
  */
 export function initAngular1(plugins) {
   if (window.angular) {
-    window.angular.module('ionic.native', []);
+    window.angular.module('ionic.native', []).run($q => {
+      window['IonicNative'].$q = $q;
+    });
 
     for (var name in plugins) {
       let serviceName = '$cordova' + name;
--- a/src/plugins/plugin.ts
+++ b/src/plugins/plugin.ts
@@ -3,7 +3,6 @@ import { Observable } from 'rxjs/Observable';
 
 declare var window;
 declare var Promise;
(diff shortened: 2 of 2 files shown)


ihadeed:
Attn. @mlynch

mlynch:
Hmm, I don't like adding a variable to IonicNative like that, it feels like the wrong solution to this issue.

Let me dig in a bit and see if there's a different way to depend on `$q`...

mattlewis92:
You're right, on more thought a better way might be to do:
```
// in plugin.ts
export const util = {};
util.Promise = Promise;
```

```
// in ng1.ts
import {util} from './plugins/plugin.ts';
window.angular.module('ionic.native', []).run($q => {
  util.Promise = $q;
});
```

This is how the [ui-router](https://github.com/angular-ui/ui-router/blob/master/src/ng1/services.ts#L221) gets around it.

mlynch:
Hey @mattlewis92, thanks for bringing this one up. I believe this is a more appropriate way to solve this as it grabs the injector from the running app properly: https://github.com/driftyco/ionic-native/commit/2dc68a4785b7d9218a0d5bb936efd35cbb554e1b

Let me know what you think.

mattlewis92:
Will this work if `$compileProvider.debugInfoEnabled(false);` is set? What if `ng-app` isn't on the body tag?

PR #459, opened 2016-08-19
Fix(File): Fixing exclusive-option (override)

DrGanmon:
Exclusive means that the file will not be overriden so it has to be set if replace is False.


ihadeed:
thanks @DrGanmon

PR #463, opened 2016-08-19
refactor(media): change MediaError to interface and error codes to static members

amuramoto:
This seems to make more sense to me. In the old implementation the getters were essentially useless, since MediaError returns a number, not the string error.

A nice side effect is this will get MediaError out of the docs TOC.


ihadeed:
thanks

Issue #439, opened 2016-08-16
writeFile issues in alternate file implementation.


ramonsan:
I am trying to save a blob in the external memory of device. I was rewriting from old ngCordova code with $cordovaFile and $cordovaFileOpener2 plugins. It correctly worked in Ionic 1 app. Now, with new methods of File implemented by  ([#283](https://github.com/driftyco/ionic-native/issues/283)) ([6ed32ef](https://github.com/driftyco/ionic-native/commit/6ed32ef)) I am trying to do using writeFile in the same app, but seems there is an issue with options parameter:

```
let storage = cordova.file.externalApplicationStorageDirectory;

  function saveFile(blob, filename) {
    return $cordovaFile.writeFile(storage, filename, blob, true);
  }
```

But get an error:

> Uncaught (in promise) TypeError: Cannot set property 'replace' of undefined(…)
>   File.writeFile @ [ionic.native.js:5015](https://github.com/driftyco/ionic-native/blob/master/src/plugins/file.ts#L653-L658)
>   [...]

OK, maybe boolean is failing, I try passing an option object:

```
  function saveFile(blob, filename) {
    return $cordovaFile.writeFile(storage, filename, blob, { replace: true });
  }
```

But another error throw:

> Uncaught (in promise) TypeError: Cannot read property 'append' of undefined(…)
>   (anonymous function) @ [ionic.native.js:5029](https://github.com/driftyco/ionic-native/blob/master/src/plugins/file.ts#L672)

OK, I will add append to object, but again same error (seems opts is not defined, but should be).

Let's do some hacking. I changed `var opts` with `var opts = {}` and try... but:

> Uncaught (in promise) FileError {code: 1, message: "NOT_FOUND_ERR"}

Finally, I try create first an empty file, and then overwrite its content:

```
  function saveFile(blob, props) {
    return $cordovaFile.createFile(storage, filename, true)
      .then(() =>
        $cordovaFile.writeFile(storage, filename, blob, { replace: true, append: true, truncate: 0 })
    );
  }
```

But got:

>   Uncaught (in promise) FileError {code: 12, message: "PATH_EXISTS_ERR"}

Some info: 

> Ionic 1.1.0
> Ionic-native: 1.3.16
> Cordova: 6.3.1

ihadeed:
@rapropos can you help us here if you have some time?

pgg:
[...]

PR #468, opened 2016-08-21
initialize writeFile options

rapropos:
intended to address [#439](https://github.com/driftyco/ionic-native/issues/439).


ihadeed:
thanks @rapropos

PR #424, opened 2016-08-15
feat(diagnostic): add full plugin functionality

ihadeed:
closes #224 

TODO:
- [x] Add constants
- [x] Proofread

PR #474, opened 2016-08-24
feat(diagnostic): added calendar methods

justinschuldt:
Added the calendar methods and switchToSettings()


jamesgroat:
Dupe of #424 ?

ihadeed:
We could use the docs, they're better than what I have in that PR. Can you push to `feat/diag`? 

Thanks!

PR #479, opened 2016-08-25
Fix(File): Fixing readFileAs

DrGanmon:
https://developer.mozilla.org/de/docs/Web/API/FileReader --> FileReader expects File and not FileEntry.


ihadeed:
Did you test it?

Also, could you please refractor the functions to arrow functions instead? thanks.

DrGanmon:
Refactoring done and yes, the code is tested in an non-public project, so unfortunately I can't upload the test. But you can easily try it on your own  :+1:

ihadeed:
thanks

PR #481, opened 2016-08-25
Fix the advanced table info

dsalvagni:
Well, seems like something was breaking the markdown table syntax.
Take a look at the screenshot below.
adv-table


ihadeed:
thanks

PR #498, opened 2016-08-27
fix(GoogleMaps): Fixes #452

### LINKED ISSUE #452: [Maps Plugin] AnimateCameraOptions - target param
Hello,

In order to be able to set map bounds on ui, AnimateCameraOptions should be able to handle a GoogleMapsLatLngBounds as well as a GoogleMapsLatLng in its target parameters.

According to cordova maps plugin example this should be possible I guess ?

Inside ionic native this could result in : 
`export interface AnimateCameraOptions {
    target?: GoogleMapsLatLng | GoogleMapsLatLngBounds;
    tilt?: number;
    zoom?: number;
    bearing?: number;
    duration?: number;
}`

ZiFFeL1992:
fixes #452

### DIFF
--- a/src/plugins/googlemaps.ts
+++ b/src/plugins/googlemaps.ts
@@ -355,7 +355,7 @@ export class GoogleMap {
  * @private
  */
 export interface AnimateCameraOptions {
-  target?: GoogleMapsLatLng;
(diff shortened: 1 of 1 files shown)


ihadeed:
thanks @ZiFFeL1992 !

PR #523, opened 2016-09-02
test: add initial test suite

mattlewis92:
This is something I really missed when contributing as I couldn't verify any of my changes without building the dist file and drop it into an ionic project which isn't a fun DX 😄 The tests are pretty basic atm but the groundwork is there to expand them. I don't think there's a lot of point writing tests for specific plugins as you'd have to mock the cordova API, but there is definitely value in testing plugin.ts + probably ng1.ts

Let me know if you want me to change anything!


ihadeed:
Thansk @mattlewis92 .. great work!

PR #529, opened 2016-09-05
fix(file): adds chunked blob writing

stalniy:
This prevents app crash when user picks a big file to write


ihadeed:
thanks @stalniy

PR #532, opened 2016-09-06
Retain Observable even during an Error Condition

Barryrowe:
The way this was setup previously, if an error occurred on the watchPosition observable, the observer was sent an error, which would have to be caught. This also has the side effect of completing the observable, which means anything down stream that would be subscribed would be unsubscribed and no longer receive updates.

Instead of using binding the error callback to ```observer.error``` this change just binds the error callback to ```observer.next``` and lets the subscriber filter out results that match ```PositionError``` rather than having to manage re-subscribing (which could just immediately fail and enter a loop of catch/retry)

### DIFF
--- a/src/plugins/geolocation.ts
+++ b/src/plugins/geolocation.ts
@@ -62,6 +62,18 @@ export interface Geoposition {
   timestamp: number;
 }
 
+export interface PositionError {
+  /**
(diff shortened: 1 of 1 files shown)


ihadeed:
thanks @Barryrowe

PR #534, opened 2016-09-08
Update background-geolocation.ts

cambronjay:
Updated documentation for BackgroundGeolocation best practices.

### DIFF
--- a/src/plugins/background-geolocation.ts
+++ b/src/plugins/background-geolocation.ts
@@ -168,7 +168,8 @@ export interface Config {
  *
  * // When device is ready :
  * platform.ready().then(() => {
- *
+ *     // IMPORTANT: BackgroundGeolocation must be called within app.ts and or before Geolocation. Otherwise the platform will not ask you for background tracking permission. 
+ * 
  *     // BackgroundGeolocation is highly configurable. See platform specific configuration options
  *     let config = {
  *             desiredAccuracy: 10,


keithdmoore:
You don't need the Geolocation plugin if you are using the Background Geolocation plugin.

cambronjay:
What if the user is currently using the app? Does BackgroundGeoLocation grab current coordinates if the app is in use? Also, if the user changes their location services settings to "While using" will that break BackgroundGeoLocation?

keithdmoore:
Looks like this wrapper is a fork of the original plugin.  In the original, you can call getCurrentPosition() at anytime.

cambronjay:
getCurrentPosition() can be called on by BackgroundGeolocation? Can you put this is in the documentation? I am seeing people use both Geolocation and BackgroundGeolocation to accomplish their goals.

keithdmoore:
It looks like this version of the plugin, that method is not available

cambronjay:
So I wonder if it is still best practice to use Geolocation with BackgroundGeolocation? Will the next version of ionic native combine the two plugins? I think clarification should be made on Geolocation versus BackgroundGeolocation and how each are affected when app permissions change. Would make more sense to combine them and have one tracking plugin. I am constantly seeing people build ionic apps using both plugins, thus the need to update the documentation about calling BackgroundGeolocation before Geolocation. If you use both and don't do this then the app will never ask for background tracking permission.

ihadeed:
@cambronjay we can't merge the two plugins together. That would be something you would need to ask the plugin developers to do. Ionic Native only provide wrappers for other plugins.

cambronjay:
@ihadeed I am talking hypothetically it would be a nice feature. Mainly this issue was about the problems that occurred when using both background geolocation and geolocation within the same application. The background geolocation plugin needs to be placed within app.ts before geolocation or it will not ask for the correct permissions. This is about just placing that note within the documentation.

PR #536, opened 2016-09-08
feat(localNotifications): added register and has permission functions

hotforfeature:
Added a couple of missing features from the cordova-plugin-local-notifications repo

https://github.com/katzer/cordova-plugin-local-notifications/wiki/04.-Scheduling#permission


ihadeed:
thanks

PR #561, opened 2016-09-20
Extension of the BackgroundGeolocation plugin

krizroring:
Extension of the BackgroundGeolocation plugin to make it compatible with the latest version of the plugin


mlynch:
Thanks!

PR #575, opened 2016-09-24
Fix exports index.ts  comments 1.3.22

ramonornela:


ramonornela:
Closes exports with `export * from <module> `

PR #576, opened 2016-09-24
Fix export isdebug

ramonornela:


mlynch:
Thanks, I had commented those out for a quick test and they're back in now so I'm going to close this. The export isDebug could be a separate PR if you're up for it, though please explain what it's doing.

PR #598, opened 2016-09-27
Fix shareWithOptions method signature

hoisel:
Fix shareWithOptions method signature, replacing 'options.file' property with 'options.files' to match social sharing plugin [method interface](https://github.com/EddyVerbruggen/SocialSharing-PhoneGap-Plugin/blob/master/src/android/nl/xservices/plugins/SocialSharing.java#L209)


ihadeed:
Thanks!

PR #599, opened 2016-09-28
feat(googlemaps): support bounds in Geocoder

zoehneto:
When using the `Geocoder.geocode` method, the area can be restricted to a rectangle using the bounds array. This array is optional and can contain two `LatLng` objects. The first `LatLng` specifies the lower left corner, the second `LatLng` specifies the upper right corner.

See:
[Implementation in the plugin](https://github.com/mapsplugin/cordova-plugin-googlemaps/blob/30a28dfee236bbdaa95e17b752f9c40116069b44/src/android/plugin/google/maps/MyGeocoder.java#L64)
[Android developer docs for the used method](https://developer.android.com/reference/android/location/Geocoder.html#getFromLocationName(java.lang.String, int, double, double, double, double))

### DIFF
--- a/src/plugins/googlemaps.ts
+++ b/src/plugins/googlemaps.ts
@@ -900,6 +900,7 @@ export class GoogleMapsLatLng {
  */
 export interface GeocoderRequest {
   address?: string;
+  bounds?: GoogleMapsLatLng[];
   position?: { lat: number; lng: number };
 }
 /**


ihadeed:
Are you sure the correct type is `GoogleMapsLatLng[]` and not `GoogleMapsLatLngBounds` ?

zoehneto:
Yes, internally this is converted to JSON and then the lat and leg properties of the object are read (I tested this change in a personal project and it works there).

ihadeed:
Alright thanks, will merge soon

martyan:
Hi Folks, is it possible to return bounds in the `GeocoderResult` result as well? 

This would be useful e.g. when you're looking for 'Amsterdam' and want to zoom to it, but you can't determine the correct zoom because you only got the `GoogleMapsLatLng`.

Cheers!

PR #600, opened 2016-09-28
fix(push): Add support for passing notification id into finish

sdsanders:
This is included in the push plugin typings but is missing from Ionic Native: https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/phonegap-plugin-push/phonegap-plugin-push.d.ts


ihadeed:
thanks

PR #603, opened 2016-09-28
fix(PowerManagement): repo url

gianpaj:
Fixes the link that is generated in the docs here
http://ionicframework.com/docs/v2/native/powermanagement/


ihadeed:
thanks

PR #611, opened 2016-10-01
chore(): deleted ionic-gulp-tslint change for gulp-tslint directly

ramonornela:
The ideal that gulp be used only in Project, It is still out of date with dependencies in package.json and with new process @ionic-app-scripts maybe o gulp be deprecated.

Thanks


ramonornela:
@ihadeed see https://github.com/driftyco/ionic-gulp-tasks/commit/e965df589e507e1ccdc8597f6ab238a978b9ea1b

Thanks

ihadeed:
thanks @ramonornela

PR #626, opened 2016-10-03
feat(zBar): add zBar barcode scanner plugin

AndreasGassmann:


src/plugins/z-bar.ts
```diff
@@ -0,0 +1,67 @@
...
+export class ZBar {
+
+  /**
+   * Open the scanner
+   * @param options { ZBarOptions } Scan options
+   * @param success { ZBarSuccessCallback } Success callback
+   * @param failure { ZBarErrorCallback } Error callback
+   */
+  @Cordova()
+  static scan(options: ZBarOptions, success: ZBarSuccessCallback, failure: ZBarErrorCallback): void { return; }
```
ihadeed: No need for successCallback and errorCallback here. 
Using the `@Cordova()` decorator with no options will automatically add these callbacks at the end of the arguments list and link them to a promise. See developer guide for more details.

So the correct line here would be:
```
static scan(options: ZBarOptions): Promise<any> { return; }
```

src/plugins/z-bar.ts
```diff
@@ -0,0 +1,67 @@
...
+
+}
+
+export interface ZBarSuccessCallback { (scanResult: string): void; }
```
ihadeed: Remove this.

src/plugins/z-bar.ts
```diff
@@ -0,0 +1,67 @@
...
+export interface ZBarErrorCallback { (error: string): void; }
```
ihadeed: Remove this.

src/plugins/z-bar.ts
```diff
@@ -0,0 +1,67 @@
...
+
+export interface ZBarOptions {
```
ihadeed: Document what each property represents here (use the information you have in the `@advanced` tag)

src/plugins/z-bar.ts
```diff
@@ -0,0 +1,67 @@
...
+ * @usage
+ * ```
+ * import { ZBar } from 'ionic-native';
+ *
+ * let zBarOptions = {
+ *       flash: "off",
+ *       drawSight: false
+ *     };
+ *
+ * ZBar.scan(zBarOptions, result => {
```
ihadeed: After fixing the method below (by returning a promise), edit the usage to something like:
```
ZBar.scan(zBarOptions).then(result => { ... }).catch(error => { ... }); 
```

AndreasGassmann:
@ihadeed Thanks for your help.

I applied your changes, but it looks like I somehow messed up the merge. If my changes are ok, I can delete my fork/pull request, create a new one and add my changes there.

ihadeed:
@AndreasGassmann yes, please do create a new clean branch and resubmit the PR. You can just copy and paste `z-bar.ts` file contents, and re-edit `index.ts` to export the plugin.

Thanks again!

PR #634, opened 2016-10-04
feat(zBar): add zBar barcode scanner plugin

AndreasGassmann:
Add support for the zBar plugin https://github.com/tjwoon/csZBar


ihadeed:
thanks @AndreasGassmann !

Bouzmine:
@AndreasGassmann I can't figure out how to make this work. When I try to use this plugin, nothing is shown and when calling a second time, the plugin rejects with error, "Scan already in progress".

PR #639, opened 2016-10-05
docs(paypal): delete ':' of the params

ramonornela:
Delete **:** of the params see http://ionic-site-staging.herokuapp.com/docs/v2/native/paypal/.


ihadeed:
thanks

PR #644, opened 2016-10-05
Changed confusing sentence structure.

thielCole:
Reading the document on which destination type, without commas leads to confusing behaviour.


ihadeed:
thanks

PR #646, opened 2016-10-06
fix Google maps setPadding not working issue

ziggyJ:
fix the issue when  sync Cordova function calls add two extra null value on the end of the args that break a lot of Cordova plugins including Google maps setPadding function

That is my simple fix, just add resolve and reject to the args if either of them is not null.
But it seems that I am touching some core functions, I am not sure if it does not break other functions.


ihadeed:
Please resubmit without the `dist` directory.

ihadeed:
Actually, we can do something like this instead:
```
if (!opts.sync) {
  args.push(resolve);
  args.push(reject);
}
```

Issue #573, opened 2016-09-23
google maps setPadding not working


arielgelbard:
google maps setPadding function is not working

Tried setPadding on a cordova project and it works no problem...

I am on ionic 2 beta 11

ramonornela:
Can provider more info of problem?

arielgelbard:
@ramonornela I call ```map.setPadding(0,0,200,0)``` and it executes and does not set the padding like the google sdk is suppose to do.

ramonornela:
What the version ionic-native?

ihadeed:
I doubt this is an issue with Ionic Native. It's probably just another conflict caused by Ionic Framework or could be a zoning issue. Try wrapping your `setPadding()` call with `ngZone.run()`.

arielgelbard:
@ramonornela the latest version

@ihadeed I tried `ngZone.run()` like this:
```
        this.map.on(GoogleMapsEvent.MAP_READY).subscribe((event) => {
            this.zone.run(() => {
                this.map.setPadding(0,0,200,0);
            });
        });
```
and is still not working for me...to confirm this, I tried `this.map.setClickable(false)` and it successfully executes no problem in `ngZone.run()`.

arielgelbard:
@ramonornela @ihadeed , have you tried setPadding and is working with an ionic 2 / ionic-native project?

uditguru:
Hey , I've also tried various workarounds but its noy working for me as well , please help with this if anyone can

ramonornela:
@uditguru Tomorrow I will try look at it 😉

ramonornela:
@uditguru I'm traveling many problems Internet since yesterday, I'll try later.

ziggyJ:
The issue is caused by ionic native add 2 extra null value on the args when the function is a sync call.

I have fixed it on my fork and I will provide a pull request.

ziggyJ:
My pull request here: https://github.com/driftyco/ionic-native/pull/646 
That is my simple fix, just add resolve and reject to the args if either of them is not null.
But it seems that I am touching some core functions,  I am not sure if it does not break other functions.

ramonornela:
Problem with one to three arguments in example code above has four arguments :wink:, more too exists problem mentioned 😉 

Thanks

ramonornela:
@uditguru the version 2.2.0 fix this?

ihadeed:
Closing this for now. We can re-open if you're still having issues.

PR #654, opened 2016-10-07
fix google maps setPadding not working issue #573

ziggyJ:
create a new pull request based on the feedback from https://github.com/driftyco/ionic-native/pull/646


ihadeed:
thanks

PR #655, opened 2016-10-07
refactor(googlemaps): throw warnings plugin_not_installed

ramonornela:


src/plugins/googlemaps.ts
```diff
@@ -1,13 +1,6 @@
...
+import { Cordova, CordovaInstance, Plugin, InstanceProperty, getPlugin, pluginWarn } from './plugin';
 import { Observable } from 'rxjs/Observable';
 
-
-/**
- * @private
- * Created by Ibrahim on 3/29/2016.
- */
-declare var plugin: any;
-
```
ramonornela: @ihadeed review this deletion, attention with copyright 😉 

Thanks

ihadeed:
Sorry I meant `@CordovaFunctionOverride` not `CordovaOverride`

ramonornela:
New decorator available in 2.0.3 I am remembering added by @mlynch https://github.com/driftyco/ionic-native/commit/f60d08b7a4110d7e183a36df5f0dadf412726ca5, later fix reports

Thanks

ihadeed:
@ramonornela sorry just had a look at the decorator again, it's not what we need here. But I was thinking we can create a decorator that will let us make sure that the plugin exists before running the function.

ramonornela:
@ihadeed this decorator works with observable that make break bc to methods returns an promise

Maybe best adjust @Cordova and @CordovaInstance accept Promise

```
@CordovaInstance({
    promiseIndex: 0, // optional if not info add push
    promise: (marker: any) {
      if (marker) {
        Promise.resolve(new GoogleMapsMarker(marker));
      } else {
        Promise.reject();
      }
    }
  })
addMarker(options: GoogleMapsMarkerOptions): Promise<GoogleMapsMarker> 
```

Thanks

ihadeed:
I think it's good enough to just have the check happen in the constructor of `GoogleMap` class and maybe the `GeoCoder` too.

ramonornela:
Maybe not throw warning but adjust to not break in browser 😉

ihadeed:
Sure we can do something like `if(!this._objectInstance) return;`

ramonornela:
@ihadeed Adjust PR, thanks by report 😉

ihadeed:
Looks good. Let's revisit this later and add utility functions in plugin.ts that return back `Promise.reject({ error: 'plugin_not_installed' });` and that observable that returns an error. So we don't have to rewrite the same thing many times.

ramonornela:
I think that this rule should applied also to called @Cordova() and @CordovaInstance() ```Promise.reject() and Observer.error()``` but in 2.2 to not generate break in 2.1.*

Maybe create a branch release 2.2 to merge improvements.

Thanks

PR #663, opened 2016-10-09
fix the name of themeablebroswer plugin, and make it works

xueron:
- The name of ThemeableBrowser was mistaked as ThemableBroswer.
- The class name and some other typos was also fix to make this plugin works fine.


ihadeed:
thanks

PR #664, opened 2016-10-09
refactor(plugin): separate decorators in other module

ramonornela:
Separation aiming at cohesion and a possible refactoring using @angular/core `TypeDecorator` to building decorators

Thanks

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -5,7 +5,7 @@ const DEVICE_READY_TIMEOUT = 2000;
 declare var window;
 
 export * from './plugins/decorators';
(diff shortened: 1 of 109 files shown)


ihadeed:
What's the benefit of this

ramonornela:
Only organizational the will to close 😉, but think rethink how they are built the decorators ```@angular/core``` help to this https://github.com/driftyco/ionic/blob/master/src/navigation/nav-util.ts#L102 and https://github.com/driftyco/ionic/blob/master/src/navigation/nav-util.ts#L112

Thanks

ihadeed:
This is already done in v3 branch, so won't be needed anymore

PR #671, opened 2016-10-11
fix(onesignal): update to match latest api

### LINKED ISSUE #667: Onesignal interfacing seems to be obsolete with Ionic 2
The current initialisation is failing with onesignal API, 
`OneSignal.init('somekey',
        { googleProjectNumber: 'somenumber', autoRegister: true })
        .subscribe();`
It fails with a promise(failure) presenting an error as unable to call apply on undefined in zone.js .
I dint dig into the code much yet but I think there is some change in the one signal API, and it must be reflected in ionic native bindings to it
tested it both on beta 11 and RC of ionic both of them fail.

**as a temporary workaround** pulling the older commit from onesignal repo works 
https://github.com/OneSignal/OneSignal-Cordova-SDK.git#ac8e1e558a425fa79695cec7482e36b05214faab

ihadeed:
closes #667

### DIFF
--- a/src/plugins/onesignal.ts
+++ b/src/plugins/onesignal.ts
@@ -14,13 +14,19 @@ import { Observable } from 'rxjs/Observable';
  * ```typescript
  * import { OneSignal } from 'ionic-native';
  *
- * OneSignal.init('b2f7f966-d8cc-11e4-bed1-df8f05be55ba',
- *                        {googleProjectNumber: '703322744261'})
- *  .subscribe(jsonData => {
- *    console.log('didReceiveRemoteNotificationCallBack: ' + JSON.stringify(jsonData));
- *  });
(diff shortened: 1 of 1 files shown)


guidux:
Hi Guys, 

I am afraid that **inFocusDisplaying** is missing in src/plugins/onesignal.ts  instead **enableNotificationsWhenActive** remains. Accordig to this [upgrade guide](https://documentation.onesignal.com/docs/upgrading-to-android-sdk-30)   from onesignal

> enableInAppAlertNotification and enableNotificationsWhenActive have been combined into inFocusDisplaying which is set from the OneSignal.Builder.

ramonornela:
Fix in https://github.com/driftyco/ionic-native/commit/babfb0dca3eb32f672062ddd8ed2b2a545c394f4

Thanks

PR #680, opened 2016-10-11
fix(themeablebrowser): add missed options

xueron:
add some missed options.


ihadeed:
thanks @xueron

xueron:
it's my pleasure :)

PR #685, opened 2016-10-12
style(file): fix angular style

ramonornela:


ihadeed:
thanks! my IDE keeps messing those up.

PR #691, opened 2016-10-13
fix(onesignal): delete method deprecated

ramonornela:
Delete enableNotificationsWhenActive() and add inFocusDisplaying() see issue https://github.com/OneSignal/OneSignal-Cordova-SDK/issues/103


ihadeed:
thanks @ramonornela !!

Issue #693, opened 2016-10-13
Expose 'type' property in GoogleMapsLatLngBounds


ColinFrick:
The LatLngBounds class of the plugin has a property named 'type' which is used to differentiate between LatLng and LatLngBounds in animateCamera and moveCamera.

Because GoogleMapsLatLngBounds doesn't expose this property animateCamera / moveCamera fails when bounds are supplied:
```
10-13 10:42:36.092: W/System.err(3895): java.lang.reflect.InvocationTargetException
10-13 10:42:36.092: W/System.err(3895): 	at java.lang.reflect.Method.invoke(Native Method)
10-13 10:42:36.092: W/System.err(3895): 	at plugin.google.maps.MyPlugin.execute(MyPlugin.java:49)
10-13 10:42:36.092: W/System.err(3895): 	at plugin.google.maps.GoogleMaps$2.run(GoogleMaps.java:259)
10-13 10:42:36.092: W/System.err(3895): 	at android.os.Handler.handleCallback(Handler.java:739)
10-13 10:42:36.092: W/System.err(3895): 	at android.os.Handler.dispatchMessage(Handler.java:95)
10-13 10:42:36.092: W/System.err(3895): 	at android.os.Looper.loop(Looper.java:148)
10-13 10:42:36.092: W/System.err(3895): 	at android.app.ActivityThread.main(ActivityThread.java:5417)
10-13 10:42:36.092: W/System.err(3895): 	at java.lang.reflect.Method.invoke(Native Method)
10-13 10:42:36.092: W/System.err(3895): 	at com.android.internal.os.ZygoteInit$MethodAndArgsCaller.run(ZygoteInit.java:726)
10-13 10:42:36.092: W/System.err(3895): 	at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:616)
10-13 10:42:36.092: W/System.err(3895): Caused by: org.json.JSONException: No value for lat
10-13 10:42:36.092: W/System.err(3895): 	at org.json.JSONObject.get(JSONObject.java:389)
10-13 10:42:36.092: W/System.err(3895): 	at org.json.JSONObject.getDouble(JSONObject.java:444)
10-13 10:42:36.092: W/System.err(3895): 	at plugin.google.maps.PluginMap.updateCameraPosition(PluginMap.java:266)
10-13 10:42:36.092: W/System.err(3895): 	at plugin.google.maps.PluginMap.animateCamera(PluginMap.java:214)
[...]

PR #694, opened 2016-10-13
fix(googlemaps): Expose 'type' property in GoogleMapsLatLngBounds #693

ColinFrick:
Adds the type property to GoogleMapsLatLngBounds

**Fixes:** #693


ihadeed:
thanks @ColinFrick

PR #699, opened 2016-10-14
fix(toast): return promise instead of observable

### LINKED ISSUE #395: Review whether Toast uses observables or Promises
Some of the methods of `Toast` use observables (for example, `show`), but usage feels more appropriate for a `Promise`. Perhaps we should revise that?

ihadeed:
closes #395


ihadeed:
Will resubmit for v3 if needed

PR #701, opened 2016-10-14
fix(google-map): destruct if callback returns more than one argument

### LINKED ISSUE #650: Can't get position for Polygon click event
In cordova-plugin-googlemaps the click event has two parameters, the first one is the clicked polygon and the second one is the clicked position:

```
polygon.on(plugin.google.maps.event.OVERLAY_CLICK, function(overlay, latLng) {
  // Handle event
});
```

But ionic-native uses an Observable that only passes through the polygon object leaving the position inaccessible.

The clicked position should be accessible in the event.

ihadeed:
fixes #650

Would this be confusing or helpful?

### DIFF
--- a/src/plugins/googlemaps.ts
+++ b/src/plugins/googlemaps.ts
@@ -126,7 +126,14 @@ export class GoogleMap {
 
     return new Observable(
       (observer) => {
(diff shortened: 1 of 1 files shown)


ihadeed:
Will resubmit for v3

PR #706, opened 2016-10-17
Correct docs on where CameraOptions constants are defined

ascorbic:
The constants used in CameraOptions are properties of Camera, not navigator.camera


ihadeed:
thanks @ascorbic !

PR #738, opened 2016-10-22
chore(scripts): clean old docs before processing

ihadeed:
This should clean the old docs before processing the new ones, to remove any old directories.


ihadeed:
@perrygovier can you have a look at this when you get a chance?

PR #745, opened 2016-10-27
docs(Transfer): add how-to download files

twaldecker:
also add information how to browse your app
files directory, which is hard to find.
I needed very long time to check where my files are.


ihadeed:
thanks

PR #749, opened 2016-10-27
fix(nfc): don't bind to name field, fix #740

### LINKED ISSUE #740: “Uncaught TypeError” while creating new ionic-native wrapper
Without any code change straight after I perform git clone, I run 
`npm install` 
`npm run build` 

Then, copied all the file under the generated dist folder to the dist folder inside my ionic-native module. The screen is all blank and I saw this from chrome debugger:

capture

ramonornela:


ihadeed:
thanks!

PR #768, opened 2016-11-03
refactor(file): add return value to be  file entry in writeFile

stalniy:
Also remove code duplication in writeExistingFile.

This is useful in cases if then you need to use `fileEntry.toURL()` for `new Trasfer().download/upload` method


mlynch:
Merging per @ihadeed's approval, thanks!

PR #769, opened 2016-11-03
docs(admob) Added basic usage description

ValentinKlinghammer:
I thought it'd be nice to give a starting point.

I'll look into documenting return types when I have some more spare time.


mlynch:
Thanks! 🍻

PR #781, opened 2016-11-09
Update background-geolocation.ts

oliviercherrier:
Update to take into account that BackgroundGeolocation.configure take only a config object in parameters and return a promise


mlynch:
This isn't correct. Our implementation exposes `configure` directly. https://github.com/driftyco/ionic-native/blob/master/src/plugins/background-geolocation.ts#L344

oliviercherrier:
Indeed. I did not use the good version off Background-geolocation.

PR #785, opened 2016-11-10
Implement EmailComposer.open()

frontweb:
This PR adds the implementation for EmailComposer.open(). It's just a Promise wrapper arround the cordova.plugins.email.open(), similar to the EmailComposer.isAvailable() method.

There's also a small change to the usage comment for EmailComposer.isAvailable() to remove the isAvailable variable since that's not sent by the promise when it resloves/rejects.

### DIFF
--- a/src/plugins/emailcomposer.ts
+++ b/src/plugins/emailcomposer.ts
@@ -17,10 +17,10 @@ declare var cordova: any;
  * import { EmailComposer } from 'ionic-native';
  *
  *
- * EmailComposer.isAvailable().then((available: boolean) =>{
- *  if(available) {
- *    //Now we know we can send
- *  }
+ * EmailComposer.isAvailable().then(() => {
+ *   // Now we know we can send
+ * }, () => {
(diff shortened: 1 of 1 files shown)


mlynch:
@frontweb, we already had the open function supported through our decorators. Was that not working? I'd rather use the decorator approach to reduce the code surface area we have to support on top of the plugin. Maybe we just had something misconfigured?

frontweb:
@mlynch you're right, the `open()` method works through the decorator, I must have done something wrong when I tried it a month ago. Sorry for the confusion.

The `isAvailable()` method documentation and/or implementation should be updated though. The current implementation uses a Promise that resolves when sending email is available and fails otherwise, so it doesn't return a boolean as mentioned in the documentation.

My PR contained a change for `isAvailable()` as well. Should I revert the changes for `open()` and leave the changes for `isAvailable()`?

frontweb:
Actually I see this has been fixed in the main repo already. I'll close this PR, thanks.

Issue #789, opened 2016-11-11
File.writeFile's options has 'replace' and 'append' inverted, breaking the function


wwoods:
Here, `ionic-native` sets `exclusive = options.replace`: https://github.com/driftyco/ionic-native/blob/master/src/plugins/file.ts#L682

And here, the Cordova plugin rejects an existing file if `exclusive` is set: https://github.com/apache/cordova-plugin-file/blob/master/src/android/LocalFilesystem.java#L119

`ionic-native` has its options flipped, should be `exclusive = !options.replace`.

Another error - the 'append' feature is broken: https://github.com/driftyco/ionic-native/blob/master/src/plugins/file.ts#L681 should have `create: !options.append`.

Other methods in this file seem to support the replace flag appropriately.

ihadeed:
There was a recent patch that was not included in your fork so I think you didn't see it yet. The current code is as follows:
```
    const getFileOpts: Flags = {
      create: !('create' in options) || options.create,
      exclusive: !!options.replace
    };
```

I'm not sure what is a better option to use, `create` or `append`. They both make sense in this situation. We can use them both and prioritize one of them. (`create: ((`create in options) && !!options.create) || !options.append)`)

As for the exclusive option, you're right. I can see here https://github.com/apache/cordova-plugin-file/blob/master/src/android/LocalFilesystem.java#L140-L145 if the file exists and we have both `create` and `exclusive` set to true, the operation will fail.

wwoods:
I actually did see that code, but it's broken, for a number of reasons.

First, with the current code, `options.replace` is still inverted.  It should be the exact opposite (still) of what it is: `exclusive = !options.replace`.  

Second, I think `{append: true}`, the old option, is much clearer than `{create: false}`.  `{create: false}` is a passive statement saying not to create a file, but it doesn't make clear what should happen instead (append?  replace bytes as they're written?).  On the other hand, if I see `File.writeFile(..., {append: true})`, I know what's happening (a new file is not being created, instead I'm expecting to append.  If the file didn't exist, I'm creating an empty file and then appending to that).

Third, and my pull request was guilty of this too (I didn't want to make unnecessary changes but rather fix what was there), I think that `WriteOptions` should have `exclusive` instead of `replace`.

[...]

PR #796, opened 2016-11-16
Fixing issue #789 - File.writeFile was broken.

wwoods:
Prior to this patch, `File.writeFile` has the `replace` and `append`
options flipped; setting `replace: true` would block the file from being
replaced, and setting `append: true` would always create a new file and
never append.

### DIFF
--- a/src/plugins/file.ts
+++ b/src/plugins/file.ts
@@ -678,8 +678,8 @@ export class File {
     }
 
     let getFileOpts: Flags = {
-      create: true,
-      exclusive: options.replace
+      create: !options.append,
+      exclusive: !options.replace
     };
 
     return File.resolveDirectoryUrl(path)


mlynch:
Thoughts on this @ihadeed? I'm not fresh on File plugin knowledge

ihadeed:
This is fixed here: https://github.com/driftyco/ionic-native/commit/867400d1ac1471d0ab7c036ca471fcce557d7281#diff-9c468d792814968d6a9ecd58630290cbR682

Your fork is outdated. Please try with the latest version of Ionic Native and report back if the issue still exists. @wwoods

ihadeed:
Just had a look at the issue you submitted here: https://github.com/driftyco/ionic-native/issues/789

You're right it does need to be fixed. Please check my comment there and maybe submit a new PR to fix the latest repo code.

PR #800, opened 2016-11-17
feat(nativegeocoder): add NativeGeocoder plugin

sebastianbaar:


mlynch:
Thanks! Will review new plugins for the next non-patch release.

ihadeed:
Thanks.

The `Cordova` decorators need to have `callbackOrder` set to `reverse`. I will edit that before publishing.

PR #804, opened 2016-11-18
Plugin reference fix

nakulgulati:
The plugin reference is available as part of the global `cordova` object and not the global `plugins` object. This caused the plugin to `Plugin not installed warning` even though the plugin is installed and works when using through `cordova.plugins.market.open(appId);`


mlynch:
Thanks, lgtm

PR #820, opened 2016-11-23
feat(onesignal) support result type OSNotificationOpenedResult

afranioce:
to handleNotificationReceived and handleNotificationOpened.


ihadeed:
I merged #888 instead since it's well documented.

PR #825, opened 2016-11-24
silence git during docs update

ihadeed:
@perrygovier when the CI tries to commit with no changes it throws a warning that gets interpreted as an error, see this build: https://circleci.com/gh/driftyco/ionic-native/1328

Thought of this quick fix, have a look when you get a chance.

### DIFF
--- a/scripts/docs/update_docs.sh
+++ b/scripts/docs/update_docs.sh
@@ -33,13 +33,13 @@ function run {
     git config --global user.email "hi@ionicframework.com"
     git config --global user.name "Ionitron"
     git add -A
-    git commit -am "Automated build of native docs driftyco/$CIRCLE_PROJECT_REPONAME@$CIRCLE_SHA1"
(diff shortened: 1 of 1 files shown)


perrygovier:
Ah, yeah, let's try a slightly different approach though, so git data is still output. Like here. 

https://github.com/driftyco/ionic/blob/master/scripts/docs/deploy.sh#L58-L75

perrygovier:
Ignore that last comment, I looked at the "Files Changed" and missed that something similar is already done higher up in the file. It's not working though. I'll investigate.

perrygovier:
I'm not 100% sure why the script thought there were changes when there were none. I'm guessing it has something to do with how bash evaluates strings for equivalency. I removed the var, putting the `git status --porcelain` in the evaluation itself. It worked this time. Let me know if you see an error again.  Either way, I think this PR covers up the problem rather than fixing it directly. It should never try to commit if there are no changes.

ihadeed:
thanks @perrygovier

PR #832, opened 2016-11-28
feat(rollbar): Add Rollbar plugin

akz92:


src/plugins/rollbar.ts
```diff
@@ -0,0 +1,37 @@
...
+export class Rollbar {
+
+  /**
+   * This function initializes the monitoring of your application
+   * @param successCallback {Function} A function to be executed when the plugin successfully initializes
+   * @param errorCallback {Function} A function to be executed when an error happens while initializing the plugin
+   * @return {Promise<any>} Returns a promise that resolves when the plugin successfully initializes
+   */
+  @Cordova()
+  static init(successCallback?: Function, errorCallback?: Function) {
```
ihadeed: Can you add a return type `: Promise<any>` to this function. Also keep `{ return; }` all in the same line.

Thanks.

akz92: @ihadeed sure. I documented this function as returning `Promise<any>` because, as far as I know, `ionic-native` wraps every method in a promise. The plugin function itself doesn't return anything. Is this the right approach or should I document it as returning `void`?

ihadeed: Here is an example of how it should look like: https://github.com/driftyco/ionic-native/blob/master/src/plugins/camera.ts#L201

The docs are fine, just need to add a TypeScript return type of `Promise<any>`.

src/plugins/rollbar.ts
```diff
@@ -0,0 +1,35 @@
...
+export class Rollbar {
+
+  /**
+   * This function initializes the monitoring of your application
+   * @param successCallback {Function} A function to be executed when the plugin successfully initializes
+   * @param errorCallback {Function} A function to be executed when an error happens while initializing the plugin
+   * @return {Promise<any>} Returns a promise that resolves when the plugin successfully initializes
+   */
+  @Cordova()
+  static init(successCallback?: Function, errorCallback?: Function): Promise<any> { return; }
```
ihadeed: You also need to remove `successCallback?: Function, errorCallback?: Function`. The `@Cordova()` decorator will automatically apply `resolve` and `reject` of the promise to these arguments.

akz92: No problem, removed it.

PR #871, opened 2016-12-12
Parameter (size) is optional

HartHerbert:
Just added the '?' to the parameter (size) in the static takePicture(size?: CameraPreviewSize): void { } because the size is a optional parameter


ihadeed:
thanks

PR #874, opened 2016-12-13
Update googlemap.ts

airasheed:
I was trying to use the Google Maps SDK as shown here [to](https://ionicframework.com/docs/v2/native/google-maps/).I followed the docs on this page verbatim. I kept getting errors and realized that i needed to wait until the platform was ready in order to access the plugin. I think by adding the platform.ready() method to the docs, it'll help newer devs implement this solution.

### DIFF
--- a/src/plugins/googlemap.ts
+++ b/src/plugins/googlemap.ts
@@ -41,6 +41,7 @@ export const GoogleMapsAnimation = {
  * @description This plugin uses the native Google Maps SDK
  * @usage
  * ```
+ * import { Platform } from 'ionic-angular';
  * import {
  *  GoogleMap,
  *  GoogleMapsEvent,
@@ -51,11 +52,14 @@ export const GoogleMapsAnimation = {
  * } from 'ionic-native';
  *
  * export class MapPage {
- *  constructor() {}
+ *  constructor(private platform: Platform) {}
  *
  * // Load map only after view is initialize
  * ngAfterViewInit() {
- *  this.loadMap();
+ *  this.platform.ready().then(()=>{
+ *    // Must wait until google maps plugin is ready
+ *    this.loadMap();
+ *  });
  * }
  *
  * loadMap() {


ihadeed:
I don't think it makes sense to add this documentation. If we do, then we have to add it for all the other plugins. We have added some documentation to show that you have to wait for `deviceready` to fire. See the `Examples` Sections on this page: http://ionicframework.com/docs/v2/native/

Gramatiik:
Sure it would make sense to add it to the docs ! So you can bear the responsibility that the docs are wrong (in a certain way) ?

ihadeed:
@Gramatiik I would rather add a general notice on all the native docs pages to remind users that you have to wait for `deviceready` to fire before calling any of the plugins.

airasheed:
@ihadeed  If you tell how you suggest the general notice be added. I undergo the responsibility to do so.

Gramatiik:
@ihadeed I agree with that method but that also means there will be "wrong code" on every plugin pages, and the notice will have to be linked on each plugin page, so it's a little bit wasting time in trying to save time :)

ihadeed:
@Gramatiik waiting for `deviceready` should be basic knowledge for Cordova developers. 

The usage part on each plugin is just a basic example showing how to use the plugin. It's not really meant to be Ionic 2 usage, just general usage. You have to remember that Ionic Native is not meant to be an Ionic 2 add on, it can be used with Ionic 1/2, any other framework, or just plain vanilla JS. Adding `platform.ready().then(() => ...)` to each doc page would be confusing for non Ionic 2 users.

We're working on improving the docs in a way that helps all users and explains everything, including the requirement to wait for `deviceready` event to fire.

airasheed:
@ihadeed Ok. Please let me know if I can help out.

Gramatiik:
You're right ! i was thinking as a Ionic user only, so if you need any help from me.
Just le me know.

PR #877, opened 2016-12-14
feat(broadcaster): add Broadcaster plugin

### LINKED ISSUE #875: Plugin request: broadcaster
From this one: https://github.com/bsorrentino/cordova-broadcaster

mebibou:
Fixes #875
plugin: https://github.com/bsorrentino/cordova-broadcaster

### DIFF
--- a/src/plugins/broadcaster.ts
+++ b/src/plugins/broadcaster.ts
@@ -1,4 +1,5 @@
 import { Plugin, Cordova } from './plugin';
+import { Observable } from 'rxjs/Observable';
 
 /**
  * @name Broadcaster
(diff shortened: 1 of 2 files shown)


src/plugins/broadcaster.ts
```diff
@@ -0,0 +1,47 @@
...
+ *
+ * ```
+ */
+@Plugin({
+  pluginName: 'Broadcaster',
+  plugin: 'cordova-plugin-broadcaster',
+  pluginRef: 'broadcaster',
+  repo: 'https://github.com/bsorrentino/cordova-broadcaster',
+  platforms: ['Android', 'iOS'],
+  install: 'ionic plugin add cordova-plugin-broadcaster'
```
ihadeed: this line is not needed since we do not have a custom install command for this plugin

src/plugins/broadcaster.ts
```diff
@@ -0,0 +1,47 @@
...
+})
+export class Broadcaster {
+
+  /**
+   * This function listen to an event sent from the native code
+   * @param eventName {string}
+   * @return {Promise<any>} Returns a promise that resolves when an event is received
+   */
+  @Cordova()
+  static addEventListener(eventName: string): Promise<any> { return; }
```
ihadeed: Might be a good idea to return an observable here.

```typescript
@Cordova({
  observable: true
})
static addEventListener(eventName: string): Observable<any> { return; }
```

mebibou: What would be the difference?

ihadeed: A promise resolves once, an observable can fire as many times as you want until you unsubscribe from it.

It should also have the following decorator instead of just `observable: true`, this will make sure we unsubscribe from the event when unsubscribing from the observable:
```
@Cordova({
  observable: true,
  clearFunction: 'removeEventListener',
  clearWithArgs: true
})
```

mebibou: ah yes correct, will do

ihadeed: No you don't need to add the `removeEventListener`. The `@Cordova` decorator will do the magic needed once you add the optioned I mentioned in my last comment.

ihadeed: Adding the `clearFunction` option will call that function when you unsubscribe from the observable. The `clearWithArgs` option will pass the same arguments passed to `addEventListener` (which are needed in this case).

PR #879, opened 2016-12-14
Update googlemaps.ts docs

Gramatiik:
Differ map loading at platform.ready() state.


ihadeed:
@Gramatiik please see my reply here: https://github.com/driftyco/ionic-native/pull/874#issuecomment-266920226

PR #880, opened 2016-12-14
Update deeplinks.ts

jguix:
The routeWithNavController method returns an observable, thus it has to be subscribed to, to get it working.

Change the documentation of the method to show that we need to subscribe to the observable, just like when using the route method.


ihadeed:
thanks

jguix:
Great!

PR #887, opened 2016-12-18
Fix #886

### LINKED ISSUE #886: Initial Network status is not available
When using the Network plugin
ionic-native/src/plugins/network.ts

My usage-
```
 Network.onConnect().subscribe(() => {
      console.log('online');
      this.online = true;
    });
```

this only fires when a network change is detected.  

There should be away to receive the network status upon app startup.  How is this accomplished?  Could a new API be added if there currently is not an elegant way to do so?

philipbrack:
Clarification in documentation that the observables do not fire right away but only when the state changes.  I suggest a way to get this information.


ihadeed:
What about the static property `type` ? You can fetch its value as soon as `deviceready` has fired.

PR #888, opened 2016-12-18
feat(onesignal): Improving OneSignal definitions

dudeofawesome:


ihadeed:
It would be nice to prefix all the interfaces with `OS` or `OneSignal` just to make sure that it doesn't conflict with anything in the future.

ihadeed:
thanks

PR #897, opened 2016-12-21
feat(speech-recognition): add new plugin

pbakondy:
Add SpeechRecognition plugin


mlynch:
Thanks! We are taking a short break for new plugins to ensure quality of existing ones but will revisit this one shortly.

PR #944, opened 2017-01-08
fix(ble): stopScan takes no args when used as clear function

### LINKED ISSUE #943: Error when unsubscribing from BLE.startScanWithOptions()
When unsubscribing from an Observable returned by `BLE.startScanWithOptions()`, which should implicitly call `BLE.stopScan()`, I get the following error on Android:
```
Error: Error: Uncaught (in promise): TypeError: callback.success.apply is not a function
TypeError: callback.success.apply is not a function
    at Object.callbackFromNative (cordova.js:293)
```
Doing some debugging in Chrome, it looks like `cordova.callbacks[callbackId]` contains
```
"{"success":[],"fail":{"reportDuplicates":true}}"
```
which happen to be the arguments as passed to `startScanWithOptions()`, and not the success/failure callbacks for `BLE.stopScan()`, as apparently expected.

Having a look at the BLE plugin source (https://github.com/driftyco/ionic-native/blob/master/src/plugins/ble.ts#L218) , the following seems suspicious:
```
 @Cordova({
    observable: true,
    clearFunction: 'stopScan',
    clearWithArgs: true
  })
static startScanWithOptions(services: string[], options: {reportDuplicates?: boolean} | any): Observable<any> { return; }
```
AFAIK, `clearWithArgs: true` would pass the arguments of the decorated function to the `clearFunction`, too. However, `BLE.stopScan()` does not expect any arguments (except for success/failure callbacks), so this is likely the cause for this.

Note that the same (`clearWithArgs: true`) also applies to `BLE.startScan()`, but I'm not using this and so I cannot verify that the same error occurs.

tkem:
Fixes #943 - `stopScan()` was called incorrectly with arguments when used as `clearFunction`.

### DIFF
--- a/src/plugins/ble.ts
+++ b/src/plugins/ble.ts
@@ -205,7 +205,7 @@ export class BLE {
   @Cordova({
     observable: true,
     clearFunction: 'stopScan',
-    clearWithArgs: true
+    clearWithArgs: false
   })
   static startScan(services: string[]): Observable<any> { return; }
 
@@ -218,7 +218,7 @@ export class BLE {
   @Cordova({
     observable: true,
     clearFunction: 'stopScan',
-    clearWithArgs: true
+    clearWithArgs: false
   })
   static startScanWithOptions(services: string[], options: {reportDuplicates?: boolean} | any): Observable<any> { return; }
 


ihadeed:
thanks @tkem

PR #946, opened 2017-01-09
fix(onesignal): Made OSNotification::app_id optional

uPaymeiFixit:
OSNotification::app_id is not required when sending a selfNotification.


ihadeed:
thanks @uPaymeiFixit

PR #949, opened 2017-01-09
Add LaunchReview

biesbjerg:


biesbjerg:
When running `ionic serve` / cordova is not available, and I access plugin method using ionic native plugin wrapper without a `catch()`, I get an error:

```
LaunchReview.launch(appId)
	.then(() => {
		console.log("Successfully launched store app");
	});
```

```
EXCEPTION: Uncaught (in promise): cordova_not_available
```

Is that expected or is the promise suppose to reject instead?

ihadeed:
@biesbjerg that's normal, Angular 2 lets you know if you didn't handle promise rejections. The promise does reject when you're running on `serve`. 

Btw, how is this plugin different than the [App Rate](http://ionicframework.com/docs/v2/native/app-rate/) plugin?

biesbjerg:
This plugin is used for launching the App Store / Google Play (and on iOS it switches to the Review tab).
It doesn't contain logic for prompting etc.

ihadeed:
@biesbjerg ok sounds good. I will merge this for the next non-patch release.

biesbjerg:
> @biesbjerg that's normal, Angular 2 lets you know if you didn't handle promise rejections. The promise does reject when you're running on serve.

The reason I'm asking is because when I access other plugins (some, not all) I get a warning in console instead of an uncaught exception:

image

Shouldn't it be handled the same?

(Ping @ihadeed - not sure you're notified of this message otherwise)

ihadeed:
@biesbjerg yeah I did a mistake a few days ago, it returns that console log when there is an uncaught exception. In `v2.2.16` it's fully fixed and it should only show  that message when the plugin isn't installed/available.

PR #966, opened 2017-01-11
feat(barcodescanner): Add missing options

### LINKED ISSUE #965: Missing options for BarcodeScanner
On Android, the `resultDisplayDuration` option from the Cordova plugin allows dismissing the scanned text display.

It may also be interesting to add `torchOn` option (Android only) that is provided by the Cordova plugin.

Should I submit a PR?

abennouna:
Fixes #965


ihadeed:
thanks @abennouna !

PR #973, opened 2017-01-16
feat(backlight): add Backlight plugin

mebibou:
plugin: https://github.com/mebibou/cordova-plugin-backlight

### DIFF
--- a/src/plugins/backlight.ts
+++ b/src/plugins/backlight.ts
@@ -1,6 +1,7 @@
 import { Plugin, Cordova } from './plugin';
 
 /**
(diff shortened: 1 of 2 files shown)


src/plugins/backlight.ts
```diff
@@ -0,0 +1,43 @@
+import { Plugin, Cordova } from './plugin';
+
+/**
+ * @name Backlight
```
ihadeed: Could you please add `@beta` tag above this line? The plugin seems very new and the performance might not be guaranteed.

PR #984, opened 2017-01-18
docs(google-maps) Make it clear that the map element needs a height

0x1ad2:
I added a change that makes it clear that the map element needs a height to be visible.

I tried to follow the docs but It didn't work, after that I thought my API keys were invalid, at the end it turned out the map wasn't visible at all.

### DIFF
--- a/src/plugins/googlemap.ts
+++ b/src/plugins/googlemap.ts
@@ -60,8 +60,9 @@ export const GoogleMapsAnimation = {
  *
  * loadMap() {
  *  // make sure to create following structure in your view.html file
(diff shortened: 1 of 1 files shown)


ihadeed:
It doesn't have to be 100%. It just needs any `height` value since the `div` default height is `0`.

0x1ad2:
@ihadeed I stated it needs *a height* to be visible, the 100% was an example. Maybe this description would be better *Be sure the HTML element where the maps get injected into has a predefined height property.*

ihadeed:
Sorry, I didn't notice that extra line of instructions. Looks good.

Issue #819, opened 2016-11-23
Add BaseClass for GoogleMap plugin


ihadeed:
BaseClass has many methods that almost every class inherits. We should add the base class into the Ionic Native wrapper, or add the methods manually to each class (worst case scenario). 

Reference: https://github.com/mapsplugin/cordova-plugin-googlemaps/wiki/BaseClass

PR #993, opened 2017-01-22
fix(google-maps): Adds some functions of BaseClass

ZiFFeL1992:
References: https://github.com/driftyco/ionic-native/issues/819

Missing functions yet:
 - removeEventListener()
 - off()
 - trigger()

### DIFF
--- a/src/plugins/googlemap.ts
+++ b/src/plugins/googlemap.ts
@@ -119,7 +119,7 @@ export class GoogleMap {
   @Cordova()
   static isAvailable(): Promise<boolean> { return; }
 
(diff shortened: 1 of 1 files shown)


ihadeed:
Thanks. I'll merge this for now for the sake of covering all the plugin's functionality. But in the future it would be nice to have a base class that all the other classes extend. I still need to do some experimenting to see how this would work with the decorators we have.

HartHerbert:
Does anyone have interest in implementing the missing functions of the BaseClass?

removeEventListener()
off()
trigger()

Issue #983, opened 2017-01-18
native Push does not include some of the phonegap-plugin-push options


uj:
Ionic v2 RC5
Some of the items in phonegap-plugin-push are not part of native push, so they seemingly cannot be used.

For Example:

push.on(...) <-- This works
push.finish(...) <-- This works

push.clearAllNotifications(...) <-- ERROR HERE: Property 'clearAllNotifications' does not exist on type 'PushNotification'

https://github.com/phonegap/phonegap-plugin-push/blob/master/docs/API.md#pushclearallnotificationssuccesshandler-errorhandler---ios--android-only

Issue #861, opened 2016-12-11
Please add support for Subscribe/Unsubscribe for Topics in Native Push


naveedahmed1:
Hi,

Sincce the plugin phonegap-plugin-push supports Subscribe/Unsubscribe for Topics it would be really nice if its support could also be addded to Ionic Native Push.

The subscribe method is used when the application wants to subscribe a new topic to receive push notifications.

https://github.com/phonegap/phonegap-plugin-push/blob/master/docs/API.md#pushsubscribetopic-successhandler-errorhandler

The unsubscribe method is used when the application no longer wants to receive push notifications from a specific topic but continue to receive other push messages.

https://github.com/phonegap/phonegap-plugin-push/blob/master/docs/API.md#pushunsubscribetopic-successhandler-errorhandler

uj:
Same goes for .clearAllNotifications(...) <-- ERROR HERE: Property 'clearAllNotifications' does not exist on type 'PushNotification'.

https://github.com/driftyco/ionic-native/issues/983

ZiFFeL1992:
You can close this as well @ihadeed :)

PR #995, opened 2017-01-22
fix(push): #983 and #861

ZiFFeL1992:
Fixes: https://github.com/driftyco/ionic-native/issues/983 and https://github.com/driftyco/ionic-native/issues/861


ihadeed:
thanks

PR #997, opened 2017-01-22
Fix exception in CordovaProperty

### LINKED ISSUE #954: Device
Im aware of the changes stated in 2.11.11 where Device.device.uuid becomes Device.uuid. 

> **BREAKING CHANGES**
> 
> device: device property no longer exists
> For example, `Device.device.platform` becomes `Device.platform now.`
> 
> https://github.com/driftyco/ionic-native/releases

**Problem description:** When calling Device.uuid in the browser, instead of getting return undefined, I get an empty object. This happens since 2.11.11

**Procedure:** `Console.log(Device.uuid);`
**Expected Outcome:** `Undefined`
**Experienced Outcome:** `{}`

peterkelly:
This reverts part of commit 67adb23, which moved the call to getPlugin()
from within CordovaProperty() to a point where it was called too early.

When the CordovaProperty decorator is called on a property, the class
containing that call has not yet had its Plugin decorator called. Only
when the latter happens does pluginRef get set. Thus, attempting to
access pluginRef within CordovaProperty at the time of the call will
give undefined, and attemping to call getPlugin with an undefined value
will throw an exception:

    Runtime Error
    Cannot read property 'split' of undefined

    TypeError: Cannot read property 'split' of undefined
        at get
        at getPlugin
        at CordovaProperty
        at __decorate
        ...

The above message can be seen in the browser after running the 'ionic
serve' command.

This commit moves the getPlugin call back to the getters and setters
established by CordovaProperty, which are only called *after* the Plugin
decorator has finished executing.

closes #954

### DIFF
--- a/src/plugins/plugin.ts
+++ b/src/plugins/plugin.ts
@@ -407,8 +407,8 @@ export function CordovaInstance(opts: any = {}) {
  * Before calling the original method, ensure Cordova and the plugin are installed.
  */
 export function CordovaProperty(target: any, key: string) {
-  const pluginInstance = getPlugin(target.pluginRef);
   const exists = () => {
+    let pluginInstance = getPlugin(target.pluginRef);
     if (!pluginInstance || pluginInstance[key] === 'undefined') {
       pluginWarn(target, key);
       return false;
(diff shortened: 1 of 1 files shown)


peterkelly:
Wrong commit

peterkelly:
You can delete this pull request; I will submit another with the correct commit

PR #998, opened 2017-01-22
Fix exception in CordovaProperty

### LINKED ISSUE #992: RunTime Error: Cannot read property 'split' of undefined
After updating to v2.3.0, running `ionic serve` causes this error. Tried this on a fresh blank project

peterkelly:
This reverts part of commit 67adb23, which moved the call to getPlugin()
from within CordovaProperty() to a point where it was called too early.

When the CordovaProperty decorator is called on a property, the class
containing that call has not yet had its Plugin decorator called. Only
when the latter happens does pluginRef get set. Thus, attempting to
access pluginRef within CordovaProperty at the time of the call will
give undefined, and attemping to call getPlugin with an undefined value
will throw an exception:

    Runtime Error
    Cannot read property 'split' of undefined

    TypeError: Cannot read property 'split' of undefined
        at get
        at getPlugin
        at CordovaProperty
        at __decorate
        ...

The above message can be seen in the browser after running the 'ionic
serve' command.

This commit moves the getPlugin call back to the getters and setters
established by CordovaProperty, which are only called *after* the Plugin
decorator has finished executing.

closes #992

### DIFF
--- a/src/plugins/plugin.ts
+++ b/src/plugins/plugin.ts
@@ -407,8 +407,8 @@ export function CordovaInstance(opts: any = {}) {
  * Before calling the original method, ensure Cordova and the plugin are installed.
  */
 export function CordovaProperty(target: any, key: string) {
-  const pluginInstance = getPlugin(target.pluginRef);
   const exists = () => {
+    let pluginInstance = getPlugin(target.pluginRef);
     if (!pluginInstance || pluginInstance[key] === 'undefined') {
       pluginWarn(target, key);
       return false;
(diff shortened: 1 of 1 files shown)


ihadeed:
Thanks

Nelly204:
Hej, i've tried to build an App with cordova. First I install Cordova@7.0.1, (this version were installed automatically when i wrote "$ npm install -g cordova") than I create a new Project and add a plattform (andoird), but when I tried  $ cordova build android> I got this Message: "Error: Cannot read property 'split' of undefined". Do you have any idea, how to fix it?
The version of Android sdk and Java jre are the current version

Thanks for answers

Issue #990, opened 2017-01-20
Plugin Request - BackgroundFetch


pdrosos:
I will most probably use [this BackgroundFetch plugin](https://github.com/transistorsoft/cordova-plugin-background-fetch) in combination with BackgroundGeolocation, to send regular location updates to server even app is in background.
It will be great if the plugin could be included in Ionic Native at some point.
https://github.com/transistorsoft/cordova-plugin-background-fetch
Thanks!

joshstrange:
Plugin added in this pull request https://github.com/driftyco/ionic-native/pull/1008

joshstrange:
Plugin is actually in this pull request now https://github.com/driftyco/ionic-native/pull/1013 due to my mistake of 2 things falling into the same pull request (which was then closed)

joshstrange:
@pdrosos This plugin has been added to the [v2.5.0 milestone](https://github.com/driftyco/ionic-native/milestone/21)

pdrosos:
@joshstrange great, thank you!

ihadeed:
This plugin was added awhile ago. Sorry I forgot to update this issue.

PR #1013, opened 2017-01-29
feat(background-fetch): Adding Background Fetch requested in #990

joshstrange:
Adding the [Background Fetch](https://github.com/transistorsoft/cordova-plugin-background-fetch) plugin to the repo.


joshstrange:
@ihadeed No rush but does the milestone need to be changed on this now that 2.5.[0|1] has gone out?

ihadeed:
@joshstrange we're taking a little break from adding new plugins. I will merge this once we're done with finalizing the v3 release.

PR #1030, opened 2017-02-03
Add FCM plugin support

rolandjitsu:


src/plugins/fcm.ts
```diff
@@ -0,0 +1,98 @@
...
+    'iOS'
+  ]
+})
+export class FCM {
+  /**
+   * Get notified every time a new token is generated.
+   * @return {Observable<any>}
+   */
+  @Cordova({
+    callbackStyle: 'node',
```
ihadeed: `callbackStyle: 'node'` is used incorrectly here. It's not needed.

`callbackStyle: 'node'` is used when the plugin expects a callback like this one:
```
function(err, result) { }
```

src/plugins/fcm.ts
```diff
@@ -0,0 +1,98 @@
...
+  static unsubscribeFromTopic(topic: string): Promise<any> {
+    return;
+  }
+
+  /**
+   * Receive push notification data.
+   * @return {Observable<any>}
+   */
+  @Cordova({
+    callbackStyle: 'node',
```
ihadeed: Remove `callbackStyle: 'node'` from here too. See comment above.

src/plugins/fcm.ts
rolandjitsu: @ihadeed ok, I thought that [this](https://github.com/fechanique/cordova-plugin-fcm/blob/master/www/FCMPlugin.js#L21) qualifies for `callbackStyle: 'node'`.

jsayol:
Just curious, is this plugin necessary when [the current Firebase plugin](https://ionicframework.com/docs/v2/native/firebase/) already offers all of FCM features and more?

ihadeed:
@jsayol yeah you're right that plugin has everything. I don't think this is needed.

PR #1031, opened 2017-02-03
feat(appodeal): adds Appodeal SDK wrappers for iOS and Android

MT--:


src/plugins/appodeal.ts
```diff
@@ -0,0 +1,443 @@
...
+   * @param interests
+   */
+  @Cordova()
+  static setInterests(interests: any): void {};
+
+  /*********************
+   * event Observables *
+   *********************/
+
+  static onInterstitialLoaded(): Observable<any> {
```
ihadeed: instead of manually writing the function, you can add the following decorator:
```typescript
@Cordova({
  eventObservable: true,
  event: 'onInterstitialLoaded'
})
```

MT--: @ihadeed Those don't work for this plugin unfortunately. :frowning:  (A previous commit has this format, and it didn't pick up the events.) I looked at other plugins that were successfully using these decorator options, but the suggested format just doesn't work. Any suggestions?

ihadeed: Do you think it's not working because the decorator listens to events on `window` instead of document?

I can modify the decorator to accept an additional property `element` that is set to `window` by default. That can be set to `document` in this case. I think other plugins can benefit from this change.

I will also use `Observable.fromEvent` instead of hard coding it.

ihadeed: Never mind, this is fine. The decorator doesn't do much here anyways. Just review the other comment I left and I think the rest is ok.

PR #1041, opened 2017-02-07
fix(plugin wrapper): fix #339, should check callback errorIndex is undefined

### LINKED ISSUE #339: Pincode dialog not working
the pincode dialog is not working. I get a promise error when running it

```
import {PinDialog} from 'ionic-native';

...

PinDialog.prompt('Enter your PIN', 'Verify PIN', ['OK', 'Cancel'])
  .then(
    (result: any) => {
      if(result.buttonIndex === 1) console.log('User clicked OK, value is: ', result.input1);
      else if(result.buttonIndex === 2) console.log('User cancelled');
    }
  );
```

### LINKED ISSUE #399: Replacing ngCordova with Ionic Native in an Ionic 1.3 Project - Geolocation Plugin
Hi,

I am having a trouble with my code. Before the adding ionic-native I had:

```
$cordovaGeolocation.watchPosition(watchOptions);
                        watch.then(
                            null,
                            function(err) {
                                watch.clearWatch();
                                $scope.resumeWatch();
                            },
                            function(position) {

                                var lat  = position.coords.latitude;
                                var long = position.coords.longitude;
                                $scope.mylocation.coords = {
                                    "latitude": lat,
                                    "longitude": long
                                };
                                console.log (lat);
                                console.log("Elapsed ->" + $scope.interval.totalMilliseconds);
                                if ($scope.interval.totalMilliseconds > 600000 || $scope.interval.totalMilliseconds == -1) {
                                    $scope.interval = MyTimer(1000);
                                    $scope.checkForNotifications(lat, long);
                                }
                            });
```

Now I have:

```
$cordovaGeolocation.watchPosition(watchOptions).subscribe(

                            function(position) {
                                $scope.$apply(function() {
                                    var lat = position.coords.latitude;
                                    var long = position.coords.longitude;
                                    $scope.mylocation.coords = {
                                        "latitude": lat,
                                        "longitude": long
                                    };
                                    console.log(lat);
                                    console.log("Elapsed ->" + $scope.interval.totalMilliseconds);
                                    if ($scope.interval.totalMilliseconds > 600000 || $scope.interval.totalMilliseconds == -1) {
                                        $scope.interval = MyTimer(1000);
                                        $scope.checkForNotifications(lat, long);
                                    }
                                });
                            },
                            function(err) {
                                watch.clearWatch();
                                $scope.resumeWatch();
                            }
                        );
```

but then I get an error:

`TypeError: $cordovaGeolocation.watchPosition(watchOptions).subscribe is not a function. (In '$cordovaGeolocation.watchPosition(watchOptions).subscribe', '$cordovaGeolocation.watchPosition(watchOptions).subscribe' is undefined)`

Do I have to include anything else beside ionic-native.js to the index file?

JiaLiPassion:
in #339, plugin such as PinDialog, the callback only have successCallback, not errorCallback, so the 
plugin wrapper in ionic-native will throw error or pass wrong parameters into native plugin because errorIndex is undefined here, https://github.com/driftyco/ionic-native/blob/master/src/plugins/plugin.ts#L179

### DIFF
--- a/src/plugins/pin-dialog.ts
+++ b/src/plugins/pin-dialog.ts
@@ -34,7 +34,8 @@ export class PinDialog {
    * @returns {Promise<{ buttonIndex: number, input1: string }>}
    */
   @Cordova({
-    successIndex: 1
+    successIndex: 1,
+    errorIndex: 4 // we don't need this, so we set it to a value higher than # of args
   })
   static prompt(message: string, title: string, buttons: string[]): Promise<{ buttonIndex: number, input1: string }> { return; }
 


ihadeed:
I'm pretty sure that in every case both `successIndex` and `errorIndex` are defined even if we don't need one of them. I don't think this is needed.

JiaLiPassion:
@ihadeed, thank you for your reply, I'll post a screenshot, pindialog plugin's API is weird, 

``` javascript
	prompt: function(message, resultCallback, title, buttonLabels) {
```

It will not throw error in errorIndex, but will insert a wrong errorHandler to prompt function.

ihadeed:
Yes, you're right. 

This is just a rare case, PinDialog and CodePush are the only plugins that have methods like this. See https://github.com/driftyco/ionic-native/blob/6ba1d4908b36d644a1de2b3bd677842ab4cb6241/src/plugins/code-push.ts#L514

I just fixed PinDialog and added `errorIndex` that's higher than the number of arguments so it's added all the way at the end. So the signature would look like:
```
  prompt(message, resultCallback, title, buttonLabels, errorHandler)
```
Now it's just an extra argument that the plugin will not use. 


It's a rare case (only 2 occurrences as of now). So, there's no need to add these lines of code to `plugin.ts` and process them for every method that has `successIndex` and `errorIndex`.

JiaLiPassion:
@ihadeed, thank you for explanation, so I just need to update https://github.com/driftyco/ionic-native/blob/6ba1d4908b36d644a1de2b3bd677842ab4cb6241/src/plugins/pin-dialog.ts to add an errorIndex = 4, is that right?

JiaLiPassion:
@ihadeed , Thank for fixing this, I will close this PR.

PR #1059, opened 2017-02-10
feat(core): auto install scripts

ihadeed:
- Adds `@ionic-native/utils` module
- Adds a new `postinstall` command to all plugins
- CLI logs/prompts when installing a plugin

Failed install:
image

User wants to modify values:
image

Successful install:
image


mlynch:
Awesome :D

A couple of thoughts:

1. Can we not use `@ionic-native/core` to keep it to just one additional non-plugin package?
2. Can you remove the prompt text "Enter value for variable:", it'll make it cleaner.

PR #1062, opened 2017-02-11
feat(inappbrowser): add interface for IAB options

ehorodyski:
One of the bigger annoyances working with the InAppBrowser is the fallibility of the options string. To mitigate that, I added a new interface, `InAppBrowserOptions`, that an object can cast itself to and be provided tooling on what keys and values are supported.

To that extent, I've overloaded the constructor and if the `options` parameter comes in as an `InAppBrowserOptions`, I map the key-value pairs into a string that is consumable by the Cordova plugin.

### DIFF
--- a/src/plugins/inappbrowser.ts
+++ b/src/plugins/inappbrowser.ts
@@ -3,6 +3,42 @@ import { Observable } from 'rxjs/Observable';
 
 declare var cordova: any;
 
+export interface InAppBrowserOptions {
+  /** Set to yes or no to ruen the InAppBrowser's location bar on or off. */
+  location?: 'yes' | 'no';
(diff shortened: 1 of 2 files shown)


ehorodyski:
Having issues debugging; will re-open when I can re-confirm on phone.

ihadeed:
As far as I know, the InAppBrowser expects features/options to be a comma separated during. Passing an object might not work.

ehorodyski:
@ihadeed -- Yeah, my changes are to give the user an interface to work with using the acceptable features, and then maps the object into a comma-separated string in the constructor. I just messed something up somewhere when I'm debugging -- maybe I didn't build my version of ionic-native correctly.

ehorodyski:
@ihadeed - I've created a new Pull Request for this with a better description and more commits.

PR #1063, opened 2017-02-11
feat(text-to-speech): Add stop method

tomsun:
https://github.com/vilic/cordova-plugin-tts/pull/32


grimace:
also need to add checkLanguage, which is in the vilic code, but does not show up in this project

PR #1065, opened 2017-02-11
feat(inappbrowser): add interface for IAB options

ehorodyski:
The InAppBrowser plugin has a lot of options/features that can be passed in during initialization, but without TypeScript it's hard to manage what options -- and their corresponding valid values -- are available.

This pull request adds the interface `InAppBrowserOptions` that will let users construct an object with tooling to display these valid options/features and suppress the values to those that are supported.

To integrate this change with the `InAppBrowser` class, I've overloaded the constructor. The `options` variable can now be of type `string` or `InAppBrowserOptions`. When the constructor is called, it type-checks the `options` parameter, if it exists. If the parameter is not a string, it will map the object into a comma-separated string that follows the InAppBrowser structure. From there it will use the mapped string inside `cordova.InAppBrowser.open()`.

For example:
    `{ hidden: "yes", location: "no" } => 'hidden=yes,location=no'`

I've provided test cases that verify the `InAppBrowser` constructors, to ensure that I am not causing conflict with the existing implementation.

**TL;DR:** This is using TypeScript to add some syntactical sugar to InAppBrowser's options. I've had many screw ups typing the options string by hand, and constantly have to go to Cordova's site for what I can actually do. I wanted to provide some tooling to make life easier.

### DIFF
--- a/src/plugins/inappbrowser.ts
+++ b/src/plugins/inappbrowser.ts
@@ -3,6 +3,42 @@ import { Observable } from 'rxjs/Observable';
 
 declare var cordova: any;
 
+export interface InAppBrowserOptions {
+  /** Set to yes or no to ruen the InAppBrowser's location bar on or off. */
+  location?: 'yes' | 'no';
+  /** Set to yes to create the browser and load the page, but not show it. The loadstop event fires when loading is complete.
+   * Omit or set to no (default) to have the browser open and load normally. */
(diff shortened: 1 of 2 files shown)


ehorodyski:
@ihadeed -- my issue yesterday when I closed the initial request? I removed and never re-added the IAB plugin -_-;

PR #1069, opened 2017-02-13
API Disclaimer - it can't be tested in a browser

dominikjenik:
Avoid users testing it in the browser.


mlynch:
Thanks, this is soon to not be true so I'm going to close (all plugins will be testable/mockable in the browser after we merge the v3 stuff)

PR #906, opened 2016-12-22
feat(app-preferences): add appPreferences plugin

thomashilzendegen:

PR #1084, opened 2017-02-16
(feat) added cordova-plugin-app-preferences support

ruffiem:
Tested


ruffiem:
Haven't seen #906 before. It seems incomplete though and I can't think how it would work.

ruffiem:
@ihadeed any idea when this will be merged ? Thank you

PR #1122, opened 2017-02-28
fix(media): add missing pauseRecord and resumeRecord methods

shanesmith:
https://github.com/apache/cordova-plugin-media#mediapauserecord
https://github.com/apache/cordova-plugin-media#mediaresumerecord

### DIFF
--- a/src/plugins/media.ts
+++ b/src/plugins/media.ts
@@ -225,6 +225,21 @@ export class MediaPlugin {
   })
   stopRecord(): void { }
 
+  /**
+   * Pauses recording
+   */
+  @CordovaInstance({
+    sync: true
+  })
+  pauseRecord(): void { }
+
+  /**
+   * Resumes recording
+   */
+  @CordovaInstance({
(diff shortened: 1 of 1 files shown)


src/plugins/media.ts
```diff
@@ -225,6 +225,21 @@ export class MediaPlugin {
   })
   stopRecord(): void { }
 
+  /**
+   * Pauses recording
+   */
+  @CordovaInstance({
+    sync: true
+  })
```
ihadeed: can you add `platforms: ['iOS']` to both methods? just below `sync: true`

shanesmith: hmm the docs seem to be out of date since Android is also supported, which i've verified on my device

https://github.com/apache/cordova-plugin-media/blob/88a69305eb7f2e7c03481f951a1d349f2a8b4ad0/src/android/AudioHandler.java#L123

shanesmith:
much appreciated!

PR #1130, opened 2017-03-02
Revert "fix(File): typing where `FileError` can be thrown"

ihadeed:
Reverts driftyco/ionic-native#1086


vegardlarsen:
Any specific reason why #1086 was reverted?

ihadeed:
@vegardlarsen the build was failing because lines like this one  https://github.com/driftyco/ionic-native/blob/master/src/plugins/file.ts#L489 were returning `Promie<FileError>`. 

I ended up implementing the changes in #1086 again, but I changed `Promise<FileError>` to `Promise<any>` so that the TypeScript compiler doesn't throw any errors. So basically your fix is currently implemented.

PR #1142, opened 2017-03-05
Syntax error resolved

musahaidari:
The `{}` should be `[]`, but since there is no binding we do not get error. In case we add one, then the syntax will be incorrect.

### DIFF
--- a/src/plugins/sqlite.ts
+++ b/src/plugins/sqlite.ts
@@ -21,7 +21,7 @@ declare var sqlitePlugin;
  * })
  *   .then((db: SQLite) => {
  *
- *     db.executeSql('create table danceMoves(name VARCHAR(32))', {}).then(() => {}).catch(() => {});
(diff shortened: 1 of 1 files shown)


src/plugins/sqlite.ts
```diff
@@ -21,7 +21,7 @@ declare var sqlitePlugin;
  * })
  *   .then((db: SQLite) => {
  *
- *     db.executeSql('create table danceMoves(name VARCHAR(32))', {}).then(() => {}).catch(() => {});
+ *     db.executeSql('create table danceMoves(name VARCHAR(32))', []).then(() => []).catch(() => {});
```
ihadeed: `.then(() => [])` should be `.then(() => {})`

ihadeed: `.then(() => [])` should be `.then(() => {})`

musahaidari: @I hameed.  Yes. You are right. Thank you for the point

musahaidari:
@Ihadeed Applied the change.

Issue #1154, opened 2017-03-08
 Paypal - 'sku' needs to be made 'optional' as explained in documented comment 


gujiman:
Hi, 
Taken from from the following issue:
https://github.com/driftyco/ionic-native/issues/1121

The variable 'sku' needs to be made 'optional' as explained in documented comment https://github.com/driftyco/ionic-native/blob/master/src/plugins/pay-pal.ts#L221:

`@param {String} sku: The stock keeping unit for this item. 50 characters max ( optional)`

**Solution would be:**

Make 'sku' optional: https://github.com/driftyco/ionic-native/blob/master/src/plugins/pay-pal.ts#L223:

`constructor(...., sku?: string) {`

Make 'sku' optional: https://github.com/driftyco/ionic-native/blob/master/src/plugins/pay-pal.ts#L249:

`sku?: string;`

PR #1161, opened 2017-03-09
Update to pay-pal.ts

gujiman:
'sku' needs to be made 'optional' as explained in the documented comment


gujiman:
As explained here:
https://github.com/driftyco/ionic-native/issues/1154

PR #1162, opened 2017-03-10
ionic2-cordova-plugin-googlemaps-v2

jalescardoso:
More information at: https://github.com/jalescardoso/ionic2-cordova-plugin-googlemaps-v2


ihadeed:
We will review this once v2 is stable. 

If you could submit a PR for the `v3-injectable` branch that would be better.

jalescardoso:
Ok, I sent it to `v3-injectable`.

According to the author will soon be stable.

Issue #810, opened 2016-11-21
[Feature request] Change In app purchase support


cleever:
The present supported plugin by ionic-native is:
https://github.com/AlexDisler/cordova-plugin-inapppurchase
However, this plugin is impossible to use, it breaks the entire app. (Android)

The purpose of Ionic Native would be a curated list of cordova plugins. 
So, in this case, it would support this plugin:
https://github.com/j3k0/cordova-plugin-purchase

It is a more complete solution and supports windows platform.

biesbjerg:
I have forked the repo and fixed the change detection bug on Android + merged a PR that fixes another bug. I think j3k0 plugin is harder to use an equally abandoned by the user, unfortunately.

My fork: https://github.com/biesbjerg/cordova-plugin-inapppurchase

mhartington:
This should be fixed in the most recent version of the plugin
It had an outdated promise polyfil.
Can you test and let me know?

mbrookson:
I've been looking into the in app purchase plugins for the past few days and as far as I can tell, the current one (https://github.com/AlexDisler/cordova-plugin-inapppurchase) supported by Ionic native is now not maintained, but also the cordova-plugin-purchase plugin (https://github.com/j3k0/cordova-plugin-purchase) seems to work better as it is the recommended way by Apple regarding listening to store changes.

I've also seen that someone has already written type definitions for it so would be good to see this support added to Ionic native as well, even if it doesn't replace the older one.

https://github.com/werk85/typed-cordova-plugin-purchase/blob/master/index.d.ts

mhartington:
Hm, yeah sounds like it might need to be added as an alternative option. 
Then at 4.0.0, we could remove the older plugin. 

@ihadeed thoughts?

ihadeed:
@mhartington 

We can definitely add the `j3k0` IAP plugin. I'll try to implement it over the next couple of weeks. 

They both seem to have tons of issues, I can't really decide which one is the ideal choice since I never used any of them. So I'm not sure if we should remove the old plugin in 4.x as of now. We can **(1)** do a community vote/survey to decide, **or** **(2)** keep both of them and document their pros/cons and leave it up to the developer to pick what they want.

ihadeed:
Hey guys,

I just implemented the plugin here: https://github.com/driftyco/ionic-native/commit/179f87b606779c05cc6f565be1883009cd430808

[...]

PR #1218, opened 2017-03-22
Update repo

expouic:
This project is no longer maintained
https://github.com/AlexDisler/cordova-plugin-inapppurchase


ihadeed:
These two plugins are not the same. We can't just simply change the repo and package ID. We will need to re-write the wrapper, or add a new one for the other plugin.

We have an issue for this here: https://github.com/driftyco/ionic-native/issues/810

Issue #1219, opened 2017-03-22
[BUG] MediaPlugin.create does not fire any callback


ramoncarreras:
I'm using ionic-native 3.1.1 and create() method of MediaPlugin class does not fire any callback.

Better explained in this [ionic forum topic](https://forum.ionicframework.com/t/mediaplayer-create-callbacks-not-fired/83805).

neoassyrian:
Yep, that's hapening with me
The Media Plugin page on ionic native has disappeared too.

ramoncarreras:
Yes. Another related topic:
https://forum.ionicframework.com/t/ionic-native-3-1-0-cordova-plugin-media-documentation/83665

PR #1220, opened 2017-03-22
fix(media-plugin): MediaPlugin.create promise never fires

vaskou:
The plugin was not working at all.

I think these are the reasons why:
1. There was a typo in MediaObject constructor
2. The MediaPlugin.create promise was never executed
3. The return object was not a MediaObject

I have tested these changes in one of my projects using startRecord, stopRecord and play functions.
I think the rest are also working. Test it before merging this pull request.


vaskou:
This is a fix for issue #1219

PR #1235, opened 2017-03-23
Camera as static non injectable service

akrisu:


mlynch:
Looks like you're on about the first example, but not the `getPicture` one. Looks like a bad auto replace on our end

ihadeed:
The constants are still instance members, they're not static.

PR #1237, opened 2017-03-23
Update index.ts

gomezcabo:
The import statement was not working properly. With the change I propose, the issue is solved.

Thanks :)


ihadeed:
The documentation is for Ionic Native 3.x. I'm guessing you're still using 2.x which uses the `ionic-native` package.

PR #1256, opened 2017-03-25
feat(intel-security): Add com-intel-security-cordova-plugin Support

ehorodyski:
This PR adds support for the [com-intel-security-cordova-plugin](https://github.com/AppSecurityApi/com-intel-security-cordova-plugin).


src/@ionic-native/plugins/intel-security/index.ts
```diff
@@ -0,0 +1,248 @@
...
+    } catch (e) {
+      console.warn('Native: Intel App Security API is not installed or you are running on a browser.');
+    }
+  }
+}
+
+/**
+ * @hidden
+ */
+export class IntelSecurityData {
```
ihadeed: You can add another `@Plugin()` tag here with a different `pluginRef` and no `repo` property. You can also make it injectable if you think that's a better way to do it.

If you make it Injectable. Then the class above should be empty and only used for documentation purposes. 

---

Apply this fix to the class below as well.

src/@ionic-native/plugins/intel-security/index.ts
```diff
@@ -0,0 +1,248 @@
...
+
+  /**
+  * This creates a new instance of secure data using plain-text data.
+  * @param id {string} Storage resource identifier.
+  * @returns {Promise<any>} Returns a Promise that resolves with the instanceID of the created data instance, or rejects with an error.
+  */
+  createFromData(data: string): Promise<any> {
```
ihadeed: After making the changes I mentioned above, you can remove the `if (!window.intel) ...` check and add the regular `@Cordova` decorator.

Apply this fix to all methods below.

ehorodyski:
Changes made, thank you!

The only thing I can't figure out is now the console is logging `cordova_not_available` from `polyfills.js`.

src/@ionic-native/plugins/intel-security/index.ts
```diff
@@ -67,17 +67,21 @@ export class IntelSecurity {
...
+  repo: ''
+})
 export class IntelSecurityData {
 
   /**
   * This creates a new instance of secure data using plain-text data.
   * @param id {string} Storage resource identifier.
   * @returns {Promise<any>} Returns a Promise that resolves with the instanceID of the created data instance, or rejects with an error.
   */
+  @Cordova()
```
ihadeed: Since now we have the Cordova decorator, there is no need to write any functionally. We can just have the keyword `return` in there.

If the real method takes an object as a parameter instead of a string, then we need to change the parameter type in the method definition/docs.

src/@ionic-native/plugins/intel-security/index.ts
```diff
@@ -4,32 +4,6 @@ import { Injectable } from '@angular/core';
 declare var window: any;
 declare var intel: any;
 
-export interface IntelSecurityDataOptions {
```
ihadeed: I liked the interface better than having the params listed below.

src/@ionic-native/plugins/intel-security/index.ts
```diff
@@ -0,0 +1,248 @@
...
+export class IntelSecurity {
+
+  public storage: IntelSecurityStorage;
+  public data: IntelSecurityData;
+
+  constructor() {
+    this.storage = new IntelSecurityStorage();
+    this.data = new IntelSecurityData();
+
+    try {
```
ihadeed: Remove these 5 lines as they are not needed.

src/@ionic-native/plugins/intel-security/index.ts
```diff
@@ -0,0 +1,248 @@
...
+}
+
+/**
+ * @hidden
+ */
+@Plugin({
+  pluginName: 'IntelSecurity',
+  plugin: 'com-intel-security-cordova-plugin',
+  pluginRef: 'intel.security.secureData',
+  repo: ''
```
ihadeed: remove the `repo` property completely.

src/@ionic-native/plugins/intel-security/index.ts
```diff
@@ -0,0 +1,248 @@
...
+ *   .then(this.intelSecurity.data.getData)
+ *   .then((data: string) => console.log(data)) // Resolves to 'Sample Data'
+ *   .catch((error: any) => console.log(error));
+ *
+ * this.intelSecurity.storage.delete(storageID)
+ *   .then(() => console.log('Deleted Successfully'))
+ *   .catch((error: any) => console.log(error));
+ *
+ * ```
+ */
```
ihadeed: Add the following
```
* @classes
* IntelSecurityData 
* IntelSecurityStorage
```

src/@ionic-native/plugins/intel-security/index.ts
```diff
@@ -0,0 +1,248 @@
...
+  * @param [options.deviceLocality] {Number} Device locality policy.
+  * @param [options.sensitivityLevel] {Number} Sensitivity level policy.
+  * @param [options.noStore] {Boolean} Disallow sealed blob access.
+  * @param [options.noRead] {Boolean} Disallow plaintext data access.
+  * @param [options.creator] {Number} Creator unique ID.
+  * @param [options.owners] {Number[]} Array of owners unique ID.
+  * @param [options.webOwners] {String[]} List of trusted web domains.
+  * @returns {Promise<any>} Returns a Promise that resolves with the instanceID of the created data instance, or rejects with an error.
+  */
+  @Cordova()
```
ihadeed: Replace all `@Cordova()` decorators with `@Cordova({ otherPromise: true })`.

I just realized their plugin returns a promise instead of using callbacks. This decorator config will convert their `Q` promise to a regular ES6 Promise.

ehorodyski:
Thanks @ihadeed -- I'm working on the changes now.

Some "author notes" from me: :-P
 * I didn't set `IntelSecurityData` or `IntelSecurityStorage` as Injectable classes because they really work in-tandem. It's my attempt at trying to keep the original `intel.security.storage` and `intel.security.data` namespacing in-tact for consumers. 
* The hesitation on adding Interfaces is based on the lack of examples and documentation from Intel on the additional (optional) properties. Some of the method documentations state that there are parameters that exist that aren't currently being used...also, some methods use object containers as a parameter, but there are a total of like 2 properties for it. Actually, I think if any method in this plugin has more than one parameter it's expected to be passed in as an object container.

Anyway, changes coming soon once I can verify they work in a sample Ionic project. Thank you for the help and guidance through this!

ehorodyski:
@ihadeed -- I'm running into this issue now after changing the `@Cordova()` decorator:

```
TypeError: Cannot read property 'constructor' of null
    at checkAvailability (main.js:81923)
    at callCordovaPlugin (main.js:82018)
    at main.js:82046
    at main.js:55921
    at new t (polyfills.js:3)
    at tryNativePromise (main.js:55920)
    at getPromise (main.js:55928)
    at wrapOtherPromise (main.js:82045)
    at main.js:82139
    at value (main.js:115414)
    at t.invoke (polyfills.js:3)
    at Object.onInvoke (main.js:35928)
    at t.invoke (polyfills.js:3)
    at e.run (polyfills.js:3)
    at polyfills.js:3
```

ihadeed:
@ehorodyski just pushed a commit to the master branch to fix this issue.

Issue #636, opened 2016-10-05
Please add cordova-plugin-zeroconf as a plugin


NeoLSN:
Please consider to add this https://github.com/becvert/cordova-plugin-zeroconf plugin in ionic-native.

ghenry22:
This would be an interesting one to allow ionic apps to find other instances of the app or other services on the network

danielsogl:
@NeoLSN you can close this issue.

danielsogl:
@ihadeed can you close it?

PR #1260, opened 2017-03-26
feat(zeroconf): add cordova-plugin-zeroconf support

tkem:
This PR adds support for [cordova-plugin-zeroconf](https://github.com/becvert/cordova-plugin-zeroconf), as requested in #636.

Please note that I only tested the browsing/watching functionality, since I've never published a zeroconf service myself (on a mobile device) and I wouldn't really know how to test that.

A note on naming: The plugin uses all lowercase `zeroconf` as pluginRef and in examples, but uses camel-case `ZeroConf` internally and sporadically in the docs. However, the [official zeroconf site](http://www.zeroconf.org/) refers to `Zeroconf` with initial caps only. I decided to go with the latter, even if that may look unfamiliar to some (but hey, so does `Geolocation`). It also avoid naming the package `zero-conf` which isn't used widely, and might cause even more confusion. 
Yes, sometimes I do pay attention to such details ;-)

### DIFF
--- a/src/@ionic-native/plugins/zeroconf/index.ts
+++ b/src/@ionic-native/plugins/zeroconf/index.ts
@@ -1,0 +1,132 @@
+import { Injectable } from '@angular/core';
+import { Cordova, Plugin } from '@ionic-native/core';
+import { Observable } from 'rxjs/Observable';
+
+export interface ZeroconfService {
+  domain: string;
+  type: string;
+  name: string;
+  port: number;
+  hostname: string;
+  ipv4Addresses: Array<string>;
+  ipv6Addresses: Array<string>;
+  txtRecord: any;
+}
+
+export interface ZeroconfResult {
+  action: 'registered' | 'added' | 'removed';
+  service: ZeroconfService;
+}
+
+/**
+ * @name Zeroconf
+ * @description
+ * This plugin allows you to browse and publish Zeroconf/Bonjour/mDNS services.
+ * @usage
+ * ```typescript
+ * import { Zeroconf } from '@ionic-native/zeroconf';
+ *
+ * constructor(private zeroconf: Zeroconf) { }
+ *
+ * ...
+ * ```
+ */
+@Plugin({
+  pluginName: 'Zeroconf',
+  plugin: 'cordova-plugin-zeroconf',
+  pluginRef: 'cordova.plugins.zeroconf',
+  repo: 'https://github.com/becvert/cordova-plugin-zeroconf'
+})
+@Injectable()
+export class Zeroconf {
+  /**
+   * Returns this device's hostname.
+   * @return {Promise<string>}
+   */
+  @Cordova()
+  getHostname(): Promise<string> { return; }
+
+  /**
+   * Publishes a new service.
+   * @usage
+   * ```typescript
+   * this.zeroconf.register('_http._tcp.', 'local.', 'Becvert\'s iPad', 80, {
+   *   'foo': 'bar'
+   * }).then(result => {
+   *   console.log('Service registered', result.service);
+   * });
+   * ```
+   * @param type {string} Service type name, e.g. "_http._tcp".
+   * @param domain {string} Domain scope of the service, typically "local.".
+   * @param name {string} Unqualified service instance name.
+   * @param port {number} Local port on which the service runs.
+   * @param txtRecord {any} Arbitrary key/value pairs describing the service.
+   * @return {Promise<ZeroconfResult>} Returns a Promise that resolves with the registered service.
+   */
+  @Cordova()
+  register(type: string, domain: string, name: string, port: number, txtRecord: any): Promise<ZeroconfResult> { return; }
+
+  /**
+   * Unregisters a service.
+   * @usage
+   * ```typescript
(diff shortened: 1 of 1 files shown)


src/@ionic-native/plugins/zeroconf/index.ts
```diff
@@ -0,0 +1,132 @@
...
+ * ```
+ */
+@Plugin({
+  pluginName: 'Zeroconf',
+  plugin: 'cordova-plugin-zeroconf',
+  pluginRef: 'cordova.plugins.zeroconf',
+  repo: 'https://github.com/becvert/cordova-plugin-zeroconf'
+})
+@Injectable()
+export class Zeroconf {
```
ihadeed: Rename the class to `ZeroConf`

src/@ionic-native/plugins/zeroconf/index.ts
```diff
@@ -0,0 +1,132 @@
...
+ * ```typescript
+ * import { Zeroconf } from '@ionic-native/zeroconf';
+ *
+ * constructor(private zeroconf: Zeroconf) { }
+ *
+ * ...
+ * ```
+ */
+@Plugin({
+  pluginName: 'Zeroconf',
```
ihadeed: Set `pluginName` to `ZeroConf`

src/@ionic-native/plugins/zeroconf/index.ts
```diff
@@ -0,0 +1,132 @@
...
+  /**
+   * Returns this device's hostname.
+   * @return {Promise<string>}
+   */
+  @Cordova()
+  getHostname(): Promise<string> { return; }
+
+  /**
+   * Publishes a new service.
+   * @usage
```
ihadeed: Move all the `@usage` documentation to the main `@usage` block in the class documentation. 

The usage here will not be rendered in the docs page.

Please apply this fix to all methods below. Simply cut & paste them above.

src/@ionic-native/plugins/zeroconf/index.ts
```diff
@@ -0,0 +1,132 @@
+import { Injectable } from '@angular/core';
+import { Cordova, Plugin } from '@ionic-native/core';
+import { Observable } from 'rxjs/Observable';
+
+export interface ZeroconfService {
```
ihadeed: Rename this to `ZeroConfService`

src/@ionic-native/plugins/zeroconf/index.ts
```diff
@@ -0,0 +1,132 @@
...
+  type: string;
+  name: string;
+  port: number;
+  hostname: string;
+  ipv4Addresses: Array<string>;
+  ipv6Addresses: Array<string>;
+  txtRecord: any;
+}
+
+export interface ZeroconfResult {
```
ihadeed: Rename this to `ZeroConfResult`

src/@ionic-native/plugins/zeroconf/index.ts
```diff
@@ -0,0 +1,132 @@
...
+  action: 'registered' | 'added' | 'removed';
+  service: ZeroconfService;
+}
+
+/**
+ * @name Zeroconf
```
ihadeed: Set the name to `ZeroConf`

src/@ionic-native/plugins/zeroconf/index.ts
tkem: Rationale, please?

src/@ionic-native/plugins/zeroconf/index.ts
tkem: I copied this style from the BLE plugin, and you're right, this isn't rendered in the docs page, too.
Do I guess this should be changed for at least the BLE plugin also, and probably other plugins are affected, too.

src/@ionic-native/plugins/zeroconf/index.ts
ihadeed: Sorry, forget about the changes regarding the capitalization of the name.

I assumed `ZeroConf` is a brand name since the plugin repo has it written in that way. I traced this thing down to this page: http://www.zeroconf.org/ and realized it's called `Zeroconf`.

I will delete the other comments below.

src/@ionic-native/plugins/zeroconf/index.ts
ihadeed: Yes this should be changed in the BLE plugin and any other similar occurrences. All usage should stay in the main `@usage` block in the class documentation.

Thanks!

src/@ionic-native/plugins/zeroconf/index.ts
tkem: Yes, it's complicated ;-)
Thanks, I really think it is easier for people who actually know/use zeroconf that way, even if "ZeroConf" looks more natural first!

ihadeed: Yes you're right.

FYI, in cases where we have brand names that need to be capitalized. We capitalize the class name but keep the package name as one word. You can see this in `OneSignal`, `PayPal`, `SQLite` .. etc

tkem: Ah, okay. Thanks, I missed that - should have looked more carefully!

PR #1289, opened 2017-03-29
chore(): use CircleCI 2

ihadeed:
Managed to drop build time to 20 seconds (from 1:45+).

- ionic-site is always cached, the scripts will take care of updating it
- node_modules gets restored from cache unless `package.json` changes (takes 10-20 more seconds to install them from scratch).
- configured it to update docs on master branch, it's a bit hacky now. Their docs say they will have a better way to do it soon

Just need to remove lines 7-9 and it's good to go. (the branches property)

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -19,12 +19,12 @@ jobs:
          paths:
          - ~/ionic-site/
       - restore_cache:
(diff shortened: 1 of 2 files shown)


perrygovier:
Wow, this is soo much faster. LTGM

Thanks!

Issue #1327, opened 2017-04-05
bug(callCordovaPlugin): Cannot read property 'apply' of undefined


AmitMY:
I get this error:
> Uncaught (in promise): TypeError: Cannot read property 'apply' of undefined
from time to time, on device.

It happens for many plugins, but this stack trace is specifically the Rollbar one.

I think it would be good to have a check before this line: https://github.com/driftyco/ionic-native/blob/9074362caea1a13923d818d6d9714bb6a254c435/src/%40ionic-native/core/plugin.ts#L130

Something like:

```
if(!pluginInstance) 
    throw new Error(`Plugin ${pluginName} has no instance`);
if(!pluginInstance[methodName]) 
    throw new Error(`Plugin ${pluginName} has no method ${methodName}`);
```

I think that in this case, as in many others, doing `console.warn` is not the best thing, as it limits the error handling capability. In my case, I use Rollbar, and thrown errors are reported to me for review.

Here is a full stack:

> Error: Uncaught (in promise): TypeError: Cannot read property 'apply' of undefined
> TypeError: Cannot read property 'apply' of undefined
> at callCordovaPlugin (http://10.0.0.15:8100/build/main.js:70204:42)
> at http://10.0.0.15:8100/build/main.js:70214:24
> at http://10.0.0.15:8100/build/main.js:49329:17
> at new t (http://10.0.0.15:8100/build/polyfills.js:3:17644)
> at tryNativePromise (http://10.0.0.15:8100/build/main.js:49328:20)
> at getPromise (http://10.0.0.15:8100/build/main.js:49336:12)
> at wrapPromise (http://10.0.0.15:8100/build/main.js:70213:93)
> at Rollbar.<anonymous> (http://10.0.0.15:8100/build/main.js:70332:20)
> at Rollbar.value [as init] (http://10.0.0.15:8100/build/main.js:122748:125)
> at http://10.0.0.15:8100/build/main.js:123032:31
> at t.invoke (http://10.0.0.15:8100/build/polyfills.js:3:14529)
> at Object.onInvoke (http://10.0.0.15:8100/build/main.js:4425:37)
> at t.invoke (http://10.0.0.15:8100/build/polyfills.js:3:14469)
> at n.run (http://10.0.0.15:8100/build/polyfills.js:3:9741)
> at http://10.0.0.15:8100/build/polyfills.js:3:6774
> 1
> File http://10.0.0.15:8100/build/main.js line 70204 col 42 in callCordovaPlugin
[...]

PR #1348, opened 2017-04-08
fix(cordova-plugin): add check if method exists

AmitMY:
For: https://github.com/driftyco/ionic-native/issues/1327

I get an error, originating there, and I need to know how to handle it

### DIFF
--- a/src/@ionic-native/core/plugin.ts
+++ b/src/@ionic-native/core/plugin.ts
@@ -32,7 +32,7 @@ export function checkAvailability(plugin: any, methodName?: string, pluginName?:
 
   pluginInstance = getPlugin(pluginRef);
 
-  if (!pluginInstance || (!!methodName && pluginInstance[methodName] === 'undefined')) {
+  if (!pluginInstance || (!!methodName && typeof pluginInstance[methodName] === 'undefined')) {
     if (!window.cordova) {
       cordovaWarn(pluginName, methodName);
       return {
@@ -54,7 +54,7 @@ export function checkAvailability(plugin: any, methodName?: string, pluginName?:
  * @private
  */
 export function instanceAvailability(pluginObj: any, methodName?: string): boolean {
-  return pluginObj._objectInstance && (!methodName || pluginObj._objectInstance[methodName] !== 'undefined');
+  return pluginObj._objectInstance && (!methodName || typeof pluginObj._objectInstance[methodName] !== 'undefined');
 }
 
 function setIndex(args: any[], opts: any = {}, resolve?: Function, reject?: Function): any {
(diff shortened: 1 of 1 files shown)


ihadeed:
See https://github.com/driftyco/ionic-native/blob/master/src/%40ionic-native/core/plugin.ts#L35

AmitMY:
Thanks @ihadeed , I did not notice it.
However, that undefined check you linked to worries me.. undefined is not triple equal to an undefined string (`undefined !== 'undefined'`)
image
So I am guessing that check is not good enough.

Also, `cordova_not_available` and `plugin_not_installed` both are not correct responses for this situation. if the plugin has an instance, but does not have a method, it should say something like `method_not_available` and print an error for the correct plugin (in console).

Do you agree?

AmitMY:
@ihadeed I managed to figure out that the plugin is the Rollbar native plugin.
This check is mandatory..

ihadeed:
Ah I see what's wrong. I forgot to add `typeof` in this line: https://github.com/driftyco/ionic-native/blob/master/src/%40ionic-native/core/plugin.ts#L35

It should be:
```
if (!pluginInstance || (!!methodName && typeof pluginInstance[methodName] === 'undefined')) {
```

Ideally all checks should be in that method and not in the one you modified. Could you please edit your PR and modify the `checkAvailability` function instead?

AmitMY:
@ihadeed Done.
The error message will be wrong though.. It will say `plugin_not_installed` instead of `method_not_available`, and will not supply any info on what method is undefined

ihadeed:
@AmitMY 

Thanks!

You're right, the message will be wrong. But I think it's good for most cases. Usually all the methods that we have in our wrappers exist on the original plugin. Which means that `methodName` should always exist.

---

Here is why we check both `methodName` and `pluginRef` and in both cases we assume that the plugin is not installed/available:

If you look at the screen orientation plugin for example, it used to add `lockOrientation` property to the `window.screen` object. In this case if we check `window.screen` only, it will exist on all browsers, but `lockOrientation` will not exists. So in this case the right error message should say "plugin not installed" or "cordova not available". 

---

If you see any plugins that might benefit from having different error messages for `pluginRef` and `methodName`, then let me know and we'll have a look at them. I don't mind having more detailed errors, but it's always good to minimize the code that we don't need. _(specially run-time stuff like this)_

---

P.S: just fixed the pluginRef for the Rollbar plugin: https://github.com/driftyco/ionic-native/commit/f3969400442dc5d0e28e14a715f3476ac139c42b

AmitMY:
@ihadeed Thanks a lot!
I guess that was the issue with the rollbar plugin.

Its funny, I have rollbar.js also, which means I have around 200 errors, 20 unique, giving me an error, because rollbar native has a wrong reference :P

I find it hilarious.

AmitMY:
Any chance for a new release? this is quite a major fix (for my code at least)

PR #1351, opened 2017-04-10
Syntax error in ShareViaEmail: typescript + ionic2

nikhilknoldus:
ShareViaEmail parameter for providing email "must be null or an array" because it's type is "string:[]" and the provided syntax is wrong, so i have updated that syntax.


nikhilknoldus:
Hi @ihadeed, pls have a look. thanks !!

PR #1359, opened 2017-04-11
Update datatypes to match recent release of plugin

aggarwalankush:
I made changes to cordova-plugin-health to accept the following array as datatypes. Earlier it was just an array of strings. Now it also contains an object with `read` and `write` array.

```
[
  'calories', 'distance',   //read and write permissions
  {
    read : ['steps'],       //read only permission
    write : ['height', 'weight']  //write only permission
  }
]
```

Don't know how to specify this array in typescript so just replaced by 'any'

### DIFF
--- a/src/@ionic-native/plugins/health/index.ts
+++ b/src/@ionic-native/plugins/health/index.ts
@@ -219,7 +219,7 @@ export class Health {
    * In Android 6 and over, this function will also ask for some dynamic permissions if needed
    * (e.g. in the case of "distance", it will need access to ACCESS_FINE_LOCATION).
    *
-   * @param {Array<any>} datatypes a list of data types you want to be granted access to.
+   * @param {(string|{ read: string[], write: string[] })[]} datatypes a list of data types you want to be granted access to.
    *
    * ```
    *  [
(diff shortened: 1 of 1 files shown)


tylervz:
The type of that array in TypeScript is `string|{ read: string[], write: string[] }[]`.

tylervz:
The type `string|{ read: string[], write: string[] }[]` would allow for
```
[
  {
    read: ['steps'],
    write: ['height', 'weight']
  }
]
```
and for
```
[
  'calories', 'distance'
]
```
but not

```
[
  'calories', 'distance',   //read and write permissions
  {
    read : ['steps'],       //read only permission
    write : ['height', 'weight']  //write only permission
  }
]
```

src/@ionic-native/plugins/health/index.ts
```diff
@@ -219,22 +219,32 @@ export class Health {
    * In Android 6 and over, this function will also ask for some dynamic permissions if needed
    * (e.g. in the case of "distance", it will need access to ACCESS_FINE_LOCATION).
    *
-   * @param {Array<String>} datatypes a list of data types you want to be granted access to
+   * @param {(string|{ read: string[], write: string[] })[]} datatypes a list of data types you want to be granted access to.
    *
+   * ```
```
ihadeed: Move this to the `@usage` block above, where ever you think makes sense

ihadeed:
Could you please do the following, after you've fixed the requested changes:

1. Define a new interface above the `Health` class:
```
export interface HealthDataType {
  /**
  * Define what this property is
  */
  read?: string[],
  /**
  * Define what this property is
  */
  write?: string[]
}
```
_(edit the docs, and I'm not sure if those are optional or not, remove the question marks if necessary)_

2. Set the types to `Array<string | HealthDataType>` instead of `(string|{ read: string[], write: string[] })[]`

aggarwalankush:
@ihadeed please review the changes.

PR #1372, opened 2017-04-14
Update secure storage plugin documentation to reflect the correct order for key:value pairs.

aurothic:
The order of the key value pair arguments were incorrect.


ihadeed:
Perhaps `set('key', 'value')` would make more sense?

aurothic:
Perhaps. I was just following the pre-established documentation. I'll push the change soon.

MTechDE:
@ihadeed can you close it?

PR #1388, opened 2017-04-18
New methods and constants available for CameraPreview

SVANNER:
Hi,
New methods related to exposure, zoom and flash modes are now available in the master.
I propose to update the wraper for Ionic, so we can use new features from Ionic.
Have performed all my tests using Ionc 2.

Kind regards


ihadeed:
awesome! thanks @SVANNER

PR #1391, opened 2017-04-18
Update package.json

MTechDE:
When users installing a v3 package they will always receive this error

```
PS C:\Users\xxx\Documents\Projekte\xxx> npm install --save @ionic-native/app-rate
ionic-hello-world@ C:\Users\xxx\Documents\Projekte\xxx
+-- UNMET PEER DEPENDENCY @angular/core@4.0.0
`-- @ionic-native/app-rate@3.4.4
```

Since Ionic 3.0 released, the package should be able to support Angular 4.0.0.
I updated the dependencies for that.

### DIFF
--- a/package.json
+++ b/package.json
@@ -4,9 +4,9 @@
   "description": "Native plugin wrappers for Cordova and Ionic with TypeScript, ES6+, Promise and Observable support",
   "license": "MIT",
   "devDependencies": {
-    "@angular/compiler": "2.4.8",
-    "@angular/compiler-cli": "2.4.8",
-    "@angular/core": "2.4.8",
+    "@angular/compiler": "4.0.0",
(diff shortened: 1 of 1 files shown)


ihadeed:
The angular dependency doesn't have a version specified.

```json
"@angular/core": "*",
```

Are you sure that warning is caused by `@ionic-native/app-rate`?

MTechDE:
I have this error with every native package I installed

ihadeed:
@MTechDE you get these warnings every time you run an `npm install` command, even if it's caused by another package. 

I'm working on a fresh Ionic 3 project right now with Ionic Native dependencies, I'm not seeing these warnings. Maybe you have an old `@ionic-native/*` package installed that is causing this warning _(older packages had a strict Angular dependency  version)_.

PR #1436, opened 2017-04-27
Update Badge plugin docs

krvajalmiguelangel:
The methods are static and should be called that way


ihadeed:
@krvajalmiguelangel I think you're still using Ionic Native 2.x... In Ionic Native 3.x, all methods are NOT static. See Badge class definition here: https://github.com/driftyco/ionic-native/blob/master/src/%40ionic-native/plugins/badge/index.ts

PR #1437, opened 2017-04-27
Type result of barcode scan

rapropos:
Make an interface type for the result of barcode scanning, so it can return something more strongly typed than `Promise<any>`.


ihadeed:
thanks @rapropos !

PR #1440, opened 2017-04-28
feat(firebase): add support for firebase ErrorLog

patrickr:
Added support for errorlogs and crash reporting in my [recent PR to cordova-plugin-firebase](https://github.com/arnesson/cordova-plugin-firebase/pull/253), this is the matching change to support that in @ionic-native/firebase


patrickr:
usage demo https://github.com/patrickr/firebaseErrorhandlerDemo

PR #1448, opened 2017-04-30
typo(docs):  Unified the documentations

MTechDE:
I know it looks like many changes, but I haven't changed that much.

- In some plugin usage parts, the "typescript" markup was missing.
- I also unified the console.log strings from console.log("hello") to console.log('Hello') so any plugin page look the same.
- And I lint some plugins.


ihadeed:
Thanks @MTechDE ! you're the best.

PR #1455, opened 2017-05-02
docs(apprate): update example

MTechDE:
Because in the example the parameter for the promptForRating() method is false, the user has to set the usesUntilPrompt option otherwise the app rate prompt will never be shown.

### DIFF
--- a/src/@ionic-native/plugins/app-rate/index.ts
+++ b/src/@ionic-native/plugins/app-rate/index.ts
@@ -113,9 +113,9 @@ export interface AppUrls {
  *
  * this.appRate.preferences.storeAppURL = {
  *   usesUntilPrompt: 3,
- *   ios: '<my_app_id>',
+ *   ios: '<app_id>',
  *   android: 'market://details?id=<package_name>',
- *   windows: 'ms-windows-store://review/?ProductId=<Store_ID>'
+ *   windows: 'ms-windows-store://review/?ProductId=<store_ID>'
  * };
  *
  * this.appRate.promptForRating(false);
--- a/src/@ionic-native/plugins/app-rate/index.ts
+++ b/src/@ionic-native/plugins/app-rate/index.ts
@@ -111,12 +111,12 @@ export interface AppUrls {
  *
  * ...
  *
- *  this.appRate.preferences.storeAppURL = {
- *    usesUntilPrompt: 3,
- *    ios: '<my_app_id>',
- *    android: 'market://details?id=<package_name>',
- *    windows: 'ms-windows-store://review/?ProductId=<Store_ID>'
- *  };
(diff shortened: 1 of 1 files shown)


ihadeed:
`usesUntilPrompt` goes in the main `preferences` object, not in `preferences.storeAppUrl`

MTechDE:
My bad. I fixed it to match a common configuration

ihadeed:
@MTechDE sorry one more change 😅 .. from my experience, things don't work as expected when overriding the `preferences` object. It's better to set the `preferences.usesUntilPrompt` and `preferences.storeAppUrl` properties separately. 

The `preferences` property is just a regular property, it doesn't have a setter/getter to take care of merging with default config. See https://github.com/pushandplay/cordova-plugin-apprate/blob/master/www/AppRate.js#L168

MTechDE:
Hm, I had problems in the past when I do it like you mentioned.

My common config looks like this

```typescript
    this.appRate.preferences = {
      openStoreInApp: true,
      displayAppName: 'Mercy',
      usesUntilPrompt: 3,
      promptAgainForEachNewVersion: true,
      storeAppURL: {
        ios: '1134880046',
        android: 'market://details?id=com.mytech.OverwatchStats'
      },
      customLocale: {
        title: "Rate Mercy",
        message: "Help us to improve your experience with Stats for Overwatch",
        cancelButtonLabel: "Pass",
        laterButtonLabel: "Ask Later",
        rateButtonLabel: "Rate it!"
      }
    };
    this.appRate.promptForRating(false);
```

So in your opinion it should look like

```typescript
 this.appRate.preferences.storeAppURL = {
   ios: '<my_app_id>',
   android: 'market://details?id=<package_name>',
   windows: 'ms-windows-store://review/?ProductId=<Store_ID>'
 };
this.appRate.preferences.usesUntilPrompt = 3;

this.appRate.promptForRating(false);
```

ihadeed:
Yeah I just looked through the source code again, it doesn't seem to matter if the properties exist or not. They're always being checked before being used. So it should be fine. But people need to know that overriding the whole `preferences` object will remove any default configuration (i.e `usesUntilPrompt`, `promptAgainForEachNewVersion`).

MTechDE:
Should I add just little note about that fact or should I use your code?

ihadeed:
Maybe use both. quick example:
```ts
// set certain preferences
this.appRate.preferences.storeAppUrl = { ... }; // ( current docs would stay here )

// or, override the whole preferences object
this.appRate.preferences = { ... } // ( your common config here, or the new example you provided )
```

MTechDE:
Okay great.
I changed it now again.

ihadeed:
thanks @MTechDE

PR #1458, opened 2017-05-03
refactor(diagnostic plugin)

dpa99c:
update camera API function signatures to support cordova.plugins.diagnostic@3.6.0. See https://github.com/dpa99c/cordova-diagnostic-plugin/commit/515317e89a874447812de0c720367c3fc68c0872

### DIFF
--- a/src/@ionic-native/plugins/diagnostic/index.ts
+++ b/src/@ionic-native/plugins/diagnostic/index.ts
@@ -622,10 +622,10 @@ export class Diagnostic extends IonicNativePlugin {
 
   /**
    * Returns the authorization status for the application to use the Camera Roll in Photos app.
-   * @returns {Promise<boolean>}
+   * @returns {Promise<string>}
    */
   @Cordova({ platforms: ['iOS'] })
-  getCameraRollAuthorizationStatus(): Promise<boolean> { return; }
+  getCameraRollAuthorizationStatus(): Promise<string> { return; }
 
   /**
    * Requests camera roll authorization for the application.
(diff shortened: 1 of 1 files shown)


ihadeed:
@dpa99c 

Thanks for the PR!

**TLDR:** It just needs one quick change, we need to specify `successIndex` and `errorIndex` in the `@Cordova()` decorator as follows:
```ts
@Cordova({
  successIndex: 1,
  errorIndex: 2
})
```

**Detailed explanation**:

When there is an optional parameter, we need to add `successIndex` and `errorIndex` properties to the `@Cordova()` decorator. The decorator can't tell what the signature of the method is, it just takes the arguments passed to the function and passes them to the original method. 

For example:
```ts
// if we call isCameraAvailable without explicitly providing value for externalStorage
this.diagnostics.isCameraAvailable();

// it will result in the original method being invoked like this
cordova.plugins.diagnostic.isCameraAvailable(onSuccess, onError);
```

So we need to specify our success and error index in the decorator as follows to fix this issue:
```ts
@Cordova({
  successIndex: 1,
  errorIndex: 0
})
```
Now the original method will be invoked with the following arguments if we use the same example above:
```js
cordova.plugins.diagnostic.isCameraAvailable(undefined, onSuccess, onError);
```

MTechDE:
We should fix some lint errors:
```
(whitespace) @ionic-native/plugins/diagnostic/index.ts[152, 38]: missing whitespace
(whitespace) @ionic-native/plugins/diagnostic/index.ts[261, 39]: missing whitespace
(whitespace) @ionic-native/plugins/diagnostic/index.ts[270, 49]: missing whitespace
(whitespace) @ionic-native/plugins/diagnostic/index.ts[279, 47]: missing whitespace
```

PR #1472, opened 2017-05-05
fix(build): fixe build errors

danielsogl:


src/@ionic-native/plugins/contacts/index.ts
```diff
@@ -79,7 +79,7 @@ export class Contact implements IContactProperties {
 
   @InstanceCheck()
   clone(): Contact {
-    let newContact = new Contact();
+    let newContact: any = new Contact();
```
ihadeed: Would it work if we set the type to `Contact`?

danielsogl: I tried it but it didn't work

danielsogl: I also had some problems with the file plugin

PR #1479, opened 2017-05-05
docs(fingerprint aio): out of beta

danielsogl:
This plugin is out of beta (1.1.0)


ihadeed:
According to the author's comment here https://github.com/NiklasMerz/cordova-plugin-fingerprint-aio/issues/16#issuecomment-291015106 the plugin seem to be ok. I guess we can remove the beta tag for this plugin.

PR #1481, opened 2017-05-05
docs(hotspot): out of beta

danielsogl:
This plugin is also out of beta since a year


ihadeed:
I think this should stay as beta. The project is discontinued https://github.com/hypery2k/cordova-hotspot-plugin#cordova-hotspot-plugin

PR #1498, opened 2017-05-07
refactor(AppPreferences): changed input type

### LINKED ISSUE #1334: [AppPreferences] store should accept string, number, boolean
as stated in the title - AppPreference-Plugin accepts `string, number, boolean` as inputs, as stated [here](https://github.com/apla/me.apla.cordova.app-preferences#notes)

currently it only accepts string as keys and value, so if I define an value (e.q. boolean -> toogle-type in app-settings.json) and want to change it in code, the saved type will be changed to string and the preference-screen will not work anymore because the type is not as expected

solution:

change

```
store(dict: string, key: string, value?: string): Promise<any>;
```

to

```
store(dict: string, key: string|number|boolean, value?: string|number|boolean): Promise<any>;
```

(key must be same because of overloading...)

danielsogl:
fix #1334

### DIFF
--- a/src/@ionic-native/plugins/app-preferences/index.ts
+++ b/src/@ionic-native/plugins/app-preferences/index.ts
@@ -46,14 +46,14 @@ export class AppPreferences extends IonicNativePlugin {
    * Set a preference value
    *
    * @param {string} dict Dictionary for key (OPTIONAL)
-   * @param {string|number|boolean} key Key
-   * @param {string|number|boolean} value Value
+   * @param {string} key Key
+   * @param {any} value Value
    * @return {Promise<any>} Returns a promise
    */
   @Cordova({
     callbackOrder: 'reverse'
   })
-  store(dict: string, key: string | number | boolean, value?: string | number | boolean): Promise<any> {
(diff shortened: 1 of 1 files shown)


src/@ionic-native/plugins/app-preferences/index.ts
```diff
@@ -46,14 +46,14 @@ export class AppPreferences extends IonicNativePlugin {
...
-   * @param {string} value Value
+   * @param {string|number|boolean} key Key
+   * @param {string|number|boolean} value Value
    * @return {Promise<any>} Returns a promise
    */
   @Cordova({
     callbackOrder: 'reverse'
   })
-  store(dict: string, key: string, value?: string): Promise<any> {
+  store(dict: string, key: string | number | boolean, value?: string | number | boolean): Promise<any> {
```
ihadeed: According to the docs it only accepts `string`. https://github.com/apla/me.apla.cordova.app-preferences/blob/master/www/apppreferences.js#L138-L139

ihadeed: Ok looks like the value can be anything, since they run it through `JSON.stringify()`. But the key has to be a string.

PR #1507, opened 2017-05-08
add args to panBy()

rapropos:
In the google-maps plugin, there is a `panBy()` function with no arguments. Looking at [the plugin code](https://github.com/mapsplugin/cordova-plugin-googlemaps/blob/master/www/googlemaps-cdv-plugin.js), it should take two numeric strings.

### DIFF
--- a/src/@ionic-native/plugins/google-maps/index.ts
+++ b/src/@ionic-native/plugins/google-maps/index.ts
@@ -363,7 +363,7 @@ export class GoogleMap {
   remove(): void { }
 
   @CordovaInstance({ sync: true })
-  panBy(): void { }
+  panBy(x: string, y: string): void { }
 }
 
 /**


src/@ionic-native/plugins/google-maps/index.ts
```diff
@@ -363,7 +363,7 @@ export class GoogleMap {
   remove(): void { }
 
   @CordovaInstance({ sync: true })
-  panBy(): void { }
+  panBy(x: string, y: string): void { }
```
topless: The parameters of panBy, are numbers since they refer to pixels.

rapropos: The plugin calls `parseInt()` on them. I believe they are strings.

PR #1536, opened 2017-05-12
feat(mix-panel): add timeEvent

nicobytes:
Add support to timeEvent for measure time duration the events


src/@ionic-native/plugins/mixpanel/index.ts
```diff
@@ -89,6 +89,17 @@ export class Mixpanel extends IonicNativePlugin {
   /**
    *
    * @param eventName {string}
+   * @returns {Promise<any>}
+   */
+  @Cordova({
```
ihadeed: There is no need for this
```ts
{
 +    successIndex: 2,
 +    errorIndex: 3
 +  }
```

PR #1550, opened 2017-05-13
chore(build): bump min core version

danielsogl:


ihadeed:
@danielsogl we only update that if the core package has a breaking/required change. As of now it makes no difference if people use 3.6.0 or 3.7.0 for the core package. It makes things simpler.

PR #1559, opened 2017-05-14
github(template): issue template

danielsogl:


.github/ISSUE_TEMPLATE.md
```diff
@@ -0,0 +1,28 @@
...
+**Related code:**
+
+```
+insert any relevant code here
+```
+
+**Other information:**
+<!-- List any other information that is relevant to your issue. Stack traces, related issues, suggestions on how to fix, Stack Overflow links, forum links, etc. -->
+
+**Ionic info:** (run `ionic info` from a terminal/cmd prompt and paste output below):
```
ihadeed: I think instead of `ionic info` we can ask for a copy of the packages installed in `package.json`. This way we know what versions of `@ionic-native/*` are they using.

danielsogl: Good point

PR #1562, opened 2017-05-15
fix(secure-storage): reject when the plugin doesn't exist

### LINKED ISSUE #1515: issue(SecureStorage) when the plugin is not installed the promises are never fulfilled
When the SecureStorage plugin is not available the function` create() `returns a `SecureStorageObject` with an `undefined` `_objectInstance`. No error thrown. If I try to call any method of the plugin (get, set, ..) I have a promise never fulfilled.  

Solution proposed

The create() method should thrown an exception for better handle this case.

eboukamza:
The `create()` method reject the promise when the plugin is not available for avoid return a never fulfilled promise when we call the methods of `SecureStorageObject`

Closes #1515

### DIFF
--- a/src/@ionic-native/plugins/secure-storage/index.ts
+++ b/src/@ionic-native/plugins/secure-storage/index.ts
@@ -8,7 +8,7 @@ declare var cordova: any;
  */
 export class SecureStorageObject {
 
-  constructor(private _objectInstance?: any) { }
+  constructor(private _objectInstance: any) { }
 
   /**
    * Gets a stored item
@@ -125,7 +125,7 @@ export class SecureStorage extends IonicNativePlugin {
       if (checkAvailability('cordova.plugins.SecureStorage', null, 'SecureStorage') === true) {
         const instance = new cordova.plugins.SecureStorage(() => res(new SecureStorageObject(instance)), rej, store);
       } else {
-        res(new SecureStorageObject());
+        rej('SecureStorage failure: cordova plugin is not available');
       }
     });
   }


ihadeed:
This behavior is intentional for plugins that create an instance of another object. In this situation the promise should ideally reject but in other cases like Google Maps it's better to create an instance object that doesn't do anything. Here's an example of why this is ideal:

```ts
let map: GoogleMap = this.googleMaps.create(...);

// the browser will throw an error because map.on doesn't exist
map.on('MAP_READY')
  .then(() => { ... });
```

I will review these plugins again and implement a universal fix. Maybe the plugins that return a promise (like this one) should reject, and the rest of them can return a dummy object.

ptitjes:
@ihadeed The problem is that the user cannot see any difference between a dummy object and a real object. So the promise should resolve to null or be rejected.

For instance, you can't do things like the following because the navigation will never be set :
```
    this.platform.ready()
      .then(() => this.secureStorage.create(SECURE_STORAGE_NAME))
      .then(storage => storage ? storage.get(SECURE_STORAGE_KEY) : null)
      .then(credentials => credentials ? this.auth.login(credentials) : false)
      .then(granted => granted ? this.nav.setRoot(MainPage) : this.nav.setRoot(LoginPage))
      .catch(error => this.nav.setRoot(LoginPage))
```

Would you mind to accept the patch from @eboukamza, please ?

eboukamza:
@ihadeed
I agree that is a quick fix but in my case that helps too to turn my app in dev mode (browser). 

If you want to work in a more elegant solution I can help you too.

src/@ionic-native/plugins/secure-storage/index.ts
```diff
@@ -125,7 +125,7 @@ export class SecureStorage extends IonicNativePlugin {
       if (checkAvailability('cordova.plugins.SecureStorage', null, 'SecureStorage') === true) {
         const instance = new cordova.plugins.SecureStorage(() => res(new SecureStorageObject(instance)), rej, store);
       } else {
-        res(new SecureStorageObject());
+        rej('SecureStorage failure: cordova plugin is not available');
```
ihadeed: Replace this block of code with:
```ts
const availabilityCheck: any = checkAvailability(SecureStorage.getPluginRef(), null, SecureStorage.getPluginName());
if (availabilityCheck === true) {
  const instance = new cordova.plugins.SecureStorage(...); // same code as above
} else {
  rej(availabilityChec.error);
}
```

ihadeed: Sorry forget about the comment above, here's a better way to do it:
```
@CordovaCheck()
create(): Promise<SecureStorageObject> {
  return new Promise<SecureStorageObject>((resolve: Function) => {
    const instance = new cordova.plugins.SecureStorage(...);
  });
}
```

The `@CordovaCheck()` decorator will make it only run if the `checkAvailability` passes, and it will automatically reject with the error if it exists.

eboukamza:
@ihadeed you are a unsolved conflict with the commit dfd0514a435b8b1a4499df99c585dbcc3058d2db in the merge 7d63e808adcfa565dccd011ffe88046eae094eca

ihadeed:
@eboukamza thanks I fixed it and published a new release now. I didn't notice that until I tried building the plugin.

PR #1580, opened 2017-05-17
docs(GoogleMaps): Moved map initialization

marcinwadon:
Hey,
The GoogleMap plugin isn't available in ngAfterViewInit() method but after deviceready event is.

Btw. there should be implemented lifecycle hook interface AfterViewInit if you wanna use ngAfterViewInit method :)

### DIFF
--- a/src/@ionic-native/plugins/google-maps/index.ts
+++ b/src/@ionic-native/plugins/google-maps/index.ts
@@ -371,6 +371,7 @@ export class GoogleMap {
  * @description This plugin uses the native Google Maps SDK
  * @usage
  * ```typescript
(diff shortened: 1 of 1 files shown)


ihadeed:
- Rule of thumb is that you have to wait for `deviceready` to fire before calling any plugin. We don't document this in every plugin's usage because it is documented in the Ionic Native's main docs page.
- The Google Maps docs show that you have to wait for `ngAfterViewInit` because we need the DOM to be ready before we pass on the HTMLElement over to the Google Maps plugin to work with it
- No need to implement `AfterViewInit`, it's optional. Refer to Angular's docs.

PR #1585, opened 2017-05-19
Worng Parameter in OnNotification()

seshachalamyv:
data.wasPressed
should be change as data.wasTapped


seshachalamyv:
Document Improvements

PR #1617, opened 2017-05-28
confirm() can return a promise that resolves to 0

nicolus:
Document the fact that confirm() will return a promise that resolves to 0 if the user clicks outside the dialog box. This behavior was previously undocummented and could lead to issues if not handled properly.


ihadeed:
Please provide a link to the original docs that mention this.

nicolus:
https://github.com/apache/cordova-plugin-dialogs#navigatornotificationconfirm

> confirmCallback: Callback to invoke with index of button pressed (1, 2, or 3) **or when the dialog is dismissed without a button press (0)**. (Function)

PR #1624, opened 2017-05-30
refactor(): marked electron ready plugins

danielsogl:
Hold your horses!
I just marked the electron/browser ready plugins.


ihadeed:
Wouldn't it be misleading to have Electron listed as a supported platform? Since there is no Cordova Electron platform really.

Correct me if in wrong, but any browser supported plugin should work on Electron too.

I think one day we will have an actual Electron platform, and plugins will have Electron specific code that taps into their APIs. That's when it would make sense to list Electron as a platform.

PR #1631, opened 2017-05-31
refactor(docs): read the description

danielsogl:
- Added missing platforms
- Unified GitHub links
- Added missing installVariables

### DIFF
--- a/src/@ionic-native/plugins/appodeal/index.ts
+++ b/src/@ionic-native/plugins/appodeal/index.ts
@@ -22,9 +22,9 @@ import { Injectable } from '@angular/core';
  */
 @Plugin({
   pluginName: 'Appodeal',
-  plugin: 'https://github.com/appodeal/appodeal-cordova-plugin',
+  plugin: 'https://github.com/appodeal/appodeal-cordova-plugin.git',
   pluginRef: 'Appodeal',
-  repo: 'https://github.com/appodeal/appodeal-cordova-plugin.git',
+  repo: 'https://github.com/appodeal/appodeal-cordova-plugin',
   platforms: ['iOS', 'Android']
 })
 @Injectable()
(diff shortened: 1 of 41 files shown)


src/@ionic-native/plugins/base64-to-gallery/index.ts
```diff
@@ -23,7 +23,7 @@ import { Cordova, Plugin, IonicNativePlugin } from '@ionic-native/core';
 @Plugin({
   pluginName: 'Base64ToGallery',
   plugin: 'cordova-base64-to-gallery',
-  pluginRef: 'cordova',
+  pluginRef: 'cordova.base64ToGallery',
```
ihadeed: This should stay as `cordova` only. 

`cordova.base64ToGallery` is a function, not an object. Our wrapper wouldn't work unless if the pluginRef is set to `cordova`.

src/@ionic-native/plugins/appodeal/index.ts
```diff
@@ -22,9 +22,9 @@ import { Injectable } from '@angular/core';
  */
 @Plugin({
   pluginName: 'Appodeal',
-  plugin: 'https://github.com/appodeal/appodeal-cordova-plugin',
+  plugin: 'https://github.com/appodeal/appodeal-cordova-plugin.git',
```
ihadeed: Any specific reason  for adding the `.git` prefix here (and other occurrences below) ?

danielsogl: I want to unify the plugins if they download the files from GitHub.
If you try to download the files without .git, GitHub redirects to the .git file.

src/@ionic-native/plugins/base64-to-gallery/index.ts
danielsogl: I will fix that

src/@ionic-native/plugins/appodeal/index.ts
danielsogl: And if you open an .git link, GitHub redirects to the repo page without .git

danielsogl: And the official Cordova docs also linking to the .git links.

PR #1638, opened 2017-06-02
docs(plugins): added platforms

danielsogl:
I read thru all plugin repositories to add or remove the supported platforms.


ihadeed:
nice!

PR #1653, opened 2017-06-04
docs(image-resizer): changed source

danielsogl:


ihadeed:
https://github.com/wymsee/cordova-imageResizer is not the same plugin

PR #1680, opened 2017-06-11
fix(package): wrong version number

danielsogl:


ihadeed:
Forgot to push my commit again. Just pushed it :)

Thanks @danielsogl

PR #1696, opened 2017-06-16
docs(clipboard): removed old plugin source

danielsogl:


EdvardGilz:
Hi, I really don't know if this is the right way to leave a comment about error issues, but the plugin doesn't do anything; when I install the plugin, inside the plugins folder, the name of the plugin is "com.danielsogl.cordova.clipboard" and when I run "ionic plugin remove danielsogl-cordova-plugin-clipboard", it shows me an error that says that the plugin is not present in the project, also in the "android", "fetch" and "ios" jsons the name that appear is "com.danielsogl.cordova.clipboard" insead something like "danielsogl-cordova-plugin-clipboard" so I think that the name is not correct, maybe is the "plugin" or "pluginRef" attribute.

ihadeed:
@EdvardGilz please open a new issue and fill out the template

PR #1705, opened 2017-06-19
(feat): change iap2 for callbacks

thielCole:
This should change the IAPProductEvents to follow the documentation and allow callbacks in their place. The documentation for the original plugin was a bit complicated. 

Didn't know what to do for the downloading callback. I haven't been able to test that piece.


ihadeed:
Thanks @thielCole

PR #1713, opened 2017-06-20
(doc): updated docs with usage example

thielCole:
Updated the doc with usage examples


thielCole:
@ihadeed Are y'all going to rename the plugin from in-app-purchase-2?

ihadeed:
I haven't gotten enough feedback regarding this plugin. Not sure if people even want it. So for now it will stay as IAP2

PR #1718, opened 2017-06-21
(doc): Improved the docs for firebase dynamic links

Stukongeluk:
The config.xml example needed to switch the ios and android tag.
Added an example how to use onDynamicLink()


Stukongeluk:
My other pull request from https://github.com/chemerisuk/cordova-plugin-firebase-dynamiclinks/pull/12 just got rejected which I also included in this one, so hold on ;)

Stukongeluk:
The config.xml preference was confusing, but after digging in deeper I found that you need the AndroidClientId inside the ios platform preference, because Firebase Invites uses this value to check if it also send the invites to Android users and vice versa.

PR #1734, opened 2017-06-26
feat(plugin): pin check

niconaso:
New plugin that allows your application to check whether pin/keyguard or passcode is setup on iOS and Android phones.


niconaso:
@ihadeed Hi! do you know if this PR could be aproved in the next release? We need to use the pin-check plugin in a project :smiley: 

Regards!
Nicolas

danielsogl:
@niconaso You can use it already without a new ionic-native release if you want to use it ^^ You already build the package so just copy the plugin to your node_modules folder

niconaso:
@danielsogl Yes, I know I can do that but we need the dependency installed on the package.json. Do you know how to specify a local npm dependency?

PR #1747, opened 2017-06-30
chore(package): bump dev deps

danielsogl:
Bump Angular, rxjs and zonejs


ihadeed:
Doesn't the framework still use angular 4.1.3?

PR #1750, opened 2017-07-01
Updated documentation to include app.module change

mohammedJ:


danielsogl:
We don't have to document this because this step is the same for every ionic-native plugin and already documented here: https://ionicframework.com/docs/native/

mohammedJ:
Ok. Not a problem. Just thought its good to have it as sometimes its
overlooked/forgotten

janpio:
Not only that @danielsogl, each plugin links to https://ionicframework.com/docs/native/#Add_Plugins_to_Your_App_Module as step 2 of its installation instructions.

But yeah, people still skip that unfortunately :(

mohammedJ:
@danielsogl , @janpio . Thanks for your time guys. I will close this pull request.

PR #1764, opened 2017-07-04
docs(readme) improve link to 2.x README

janpio:
link to branch instead of specific version so possible future updates are reflected after clicking the link


ihadeed:
Isn't AngularJS the same as Angular 1.x ? Maybe we should just keep AngularJS there

janpio:
It is, but I thought maybe saying "Angular 1" was a conscious decision here to make sure people find what they are looking for. Some still refer to AngularJS as Angular and maybe get the hint via "v1".

PR #1780, opened 2017-07-09
Fix parameter in the trackMetric method

victorsosa:
In the trackMetric the first and second parameter is a number type.


ihadeed:
Please provide a link to the original plugin that supports this change.

victorsosa:
https://github.com/danwilson/google-analytics-plugin/blob/master/typescript/analytics.d.ts#L35

ihadeed:
thanks !

PR #1799, opened 2017-07-13
Use RxJS 5.4.2 to fix generics checking in TS 2.4.1

GulinSS:
Fix for https://github.com/ionic-team/ionic-native/issues/1798

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -4624,9 +4624,9 @@
       "dev": true
     },
     "rxjs": {
-      "version": "5.1.1",
-      "resolved": "https://registry.npmjs.org/rxjs/-/rxjs-5.1.1.tgz",
-      "integrity": "sha1-/EiSKWW8bF77zA/kbpCjr2QTens=",
+      "version": "5.4.2",
+      "resolved": "https://registry.npmjs.org/rxjs/-/rxjs-5.4.2.tgz",
+      "integrity": "sha1-KjI2/L8D31e64G/Wly/ZnlwI/Pc=",
       "dev": true
     },
     "safe-buffer": {
--- a/package.json
+++ b/package.json
@@ -39,7 +39,7 @@
     "q": "1.5.0",
     "queue": "4.2.1",
     "rimraf": "2.6.1",
-    "rxjs": "5.4.0",
+    "rxjs": "5.4.2",
     "semver": "5.3.0",
     "tslint": "3.15.1",
     "tslint-ionic-rules": "0.0.8",


timbru31:
What's the status of this? This is only a bugfix release according to semver and their [changelog](https://github.com/ReactiveX/rxjs/blob/master/CHANGELOG.md), but brings support for TypeScript 2.4, which is awesome 👍   
Besides that, @angluar/cli v1.3.0 is right around the corner with support for TS 2.4, too.

timbru31:
@GulinSS could you even update to RxJS v5.4.3? That one came out in the meantime and adds support for TypeScript v2.4.2.

`@angular/cli` v1.3.0 is now live, so this is currently our only blocking issue. Would be awesome to see this merged and released soon!

ihadeed:
If they update the Rxjs version in the framework/app-base repo, then we can update it here. We're keeping the dependencies as consistent as possible. 

https://github.com/ionic-team/ionic2-app-base/blob/master/package.json#L29

mbielski:
TypeScript is now at 2.5 and this module being stuck at 2.3 is preventing me from updating my project. It looks like the only merge conflict is the package.json file. Can somebody please resolve this?

timbru31:
@mbielski No they simply to refuse to pull this in, since the "ionic2-app-base" project needs to be updated first. My PR is open since >1 week now without any feedback.
And here is another PR that's not being merged since months: https://github.com/ionic-team/ionic/pull/12000

danielsogl:
They have to merge this PR first: https://github.com/ionic-team/ionic/pull/12000

danielsogl:
@timbru31 Did I hear my name? 💃

mlynch:
We didn't refuse to merge it, we're trying to be careful about dependency changes that can break things. It's going to take longer than a week to merge PRs given all the moving parts. We appreciate the contribution and understanding

mbielski:
I can understand a week or two, even three. This is 8 days short of 2 months. The version of TypeScript proposed in the PR isn't even the current version anymore. I don't expect all projects to stay on current versions of all dependencies, but staying updated on current versions of core things like TypeScript seem to me like they should be a much higher priority.

mlynch:
Unfortunately, even typescript updates can cause unforeseen issues and staying on the latest release of any dependency is not our number one priority. Rather, stability is. We prioritize staying in sync with Ionic 2 app base, which is at 2.3.4. We will adjust this soon, once we have the chance to test and confirm there aren't any issues.

timbru31:
https://github.com/ionic-team/ionic2-app-base has been updated with rxjs 5.4.3 (via https://github.com/ionic-team/ionic2-app-base/commit/fb3a00827f6252064b7340b4afcc7a05849e2530) - can we update this PR please? It's blocking our TS update to 2.4+ since months.

timbru31:
#2036 does the same.

PR #1801, opened 2017-07-13
Change install to use cordova-plugin-googlemaps

MrPutuLips:
Plugin is available via https://www.npmjs.com/package/cordova-plugin-googlemaps


ihadeed:
That's v1 .. the new wrapper in ionic native 4.0 supports v2 which is still in beta. Should be released as stable soon.

PR #1828, opened 2017-07-19
Fix errors on ionic google maps docs

iget-master:
I've spent hours trying to just add and google map on my application, and always the application was closing with the exact same example of docs.

The problem is that the example is incoherent: It says that you must wait the map ready event prior to do anything with the map, but the example itself do the opposite.

I've updated the docs to put the code in the right place, inside the event listener callback.

Also I've removed the let `const marker: Marker` piece, since `map.addMarker()` method returns a `Promise<Marker>`, not a `Marker`, so the code also fails.

This kind of incoherence on Ionic docs make many developers crazy and about stop using the platform, so you guys should take care about it, **put docs in first place**!


ihadeed:
Thanks. @wf9a5m75 updated this in #1834

PR #1834, opened 2017-07-24
Fix: Google Maps Doc page

wf9a5m75:


wf9a5m75:
@ihadeed Could you update the plugin?

And how to install the @ionic-native/googlemaps plugin from local.
Since I can not install the plugin from local, I can not test the code.

ihadeed:
@wf9a5m75 

I'll take a look and publish it soon.

To install locally, you can run `npm run build google-maps` in the root directory of `ionic-native`. Then copy `dist/google-maps` directory and use it instead of the old `node_modules/@ionic-native/google-maps` in your Ionic app.

You can also use `npm link` but I'm not quite sure how that works.

wf9a5m75:
Ok, thank you. I will try it tomorrow

wf9a5m75:
I tried `npm run build google-maps` , and it worked!
(I could not wait tomorrow)

I will test all methods in couple of days, so could you wait for a while?

----

By the way, I also tried `npm link`, but it caused `@ionic-native/core` module is not found, even it was installed. I just let you know.

wf9a5m75:
I tested couple of methods, and I noticed there are some incorrect code. I will test all methods one by one. Please wait for a while. I will let you know when I check all methods.

src/@ionic-native/plugins/google-maps/index.ts
```diff
@@ -540,30 +638,18 @@ export class GoogleMaps extends IonicNativePlugin {
   pluginRef: 'plugin.google.maps.BaseArrayClass',
   repo: ''
 })
-export class BaseArrayClass<T> extends IonicNativePlugin {
-  private _objectInstance: any;
+export class BaseArrayClass<T> extends BaseClass {
 
-  constructor(initialData: T[]) {
+  constructor(initialData?: any) {
```
ihadeed: Perhaps we could use `: T[] | any` to indicate that we can pass an array of the same type, or `any` (BaseArrayClass instance).

wf9a5m75: Ah, you are right.

wf9a5m75: I will change it later.

wf9a5m75:
By the way, the plugin contributers and I are creating a demo app that demonstrate all features with example code.

You can use this demo app for test when you need to update the wrapper plugin.
https://github.com/wf9a5m75/ionic-googlemaps-demo

ihadeed:
@wf9a5m75 

Sounds good.

Also, checkout a small demo that I made here:
https://github.com/zyra/ionic-native-playground/blob/master/src/components/google-map/google-map.ts
https://github.com/zyra/ionic-native-playground/tree/master/src/pages/google-maps

I was hoping to create a modular google maps component that makes it super easy to use google maps anywhere, instead of setting everything up manually. 

I also thought about creating a hybrid Google Maps component/module that wraps your plugin and the JavaScript SDK. This way we would have a high performance Google Maps that works on both mobile apps and PWAs. I think I have some code on my laptop, can't remember how developed it is though.

wf9a5m75:
>I also thought about creating a hybrid Google Maps component/module that wraps your plugin and the JavaScript SDK. 

Is this going to be official component, or just playground?

Because many people come to the cordova-googlemaps-plugin repository.
I don't want to receive the issues of not my plugin.

ihadeed:
@wf9a5m75 

It's a module, but not part of Ionic Native. It's not within the scope of this project.

wf9a5m75:
Okay, thank you for clarify. I'm relieved.
I don't support the browser platform yet.

By the way, I think `Observable.fromEvent(this._objectInstance, eventName);` has a problem.
It seems this drops the second argument of callback.

In the js plugin, if I do `marker.trigger(position, marker)`, the event listener has to receive both.
But the ionic-wrapper version drops the second argument `marker`.

wf9a5m75:
I found the solution.
https://github.com/ReactiveX/rxjs/issues/2431#issuecomment-283364650

ihadeed:
@wf9a5m75 observables/promises can only pass 1 argument. You must convert them to an array/object.

wf9a5m75:
Yes, I tested so deeply. But this is hard choice.
Because the JS plugin already passes two arguments, but ionic accepts only one.
Converting array/object both create different code from the JS version ... :(

wf9a5m75:
What I wanted thing was:

```js
this.map.addMarker({...}).then(onMarkerAdded);
this.map.addMarker({...}).then(onMarkerAdded);
this.map.addMarker({...}).then(onMarkerAdded);

onMarkerAdded(marker) => {
  marker.on(GoogleMapsEvent.MARKER_CLICK, this.onMarkerClick);
}

onMarkerClick(latLng: ILatLng, marker: Marker) {
  marker.showInfoWindow();
}
```

Using common event listener saves memory usage.
But ionic (typescript?) does not allow this.

If I pass the marker instance as the first argument, you can do this.
But unfortunately, all marker events passes their position.
Because the regular JS can get the marker instance from `this`.
So there was no reason to pass the marker instance to argument, and I also designed events are similar with the Google Maps JavaScript API v3.
That's why the plugin passes the marker position.

```js
onMarkerClick(latLng) {
  var marker = this;
  marker.showInfoWindow();
}
```

-------

If this time were one year ago (or half year ago), I change the argument from position to marker instance.
But there are already lots of users.
So I don't want to change the definition anymore.

I noticed this today(yesterday), and I spent almost 7 or 8 hours. Then I finally gave up this way.

ihadeed:
@wf9a5m75 

I tried addressing this problem a long time ago. 

https://github.com/ionic-team/ionic-native/pull/701/files

This basically converts the arguments to an array in case there was only one argument passed.

wf9a5m75:
Yes, I also archived the observable passes the multiple arguments. Almost the same as you did.
The problem is `Promise.resolve()`.
Even if I pass the multiple arguments to the `promise.resolve()`, it accepts only the first arguments.

ihadeed:
@wf9a5m75 
I have a fix for that too here: https://github.com/ionic-team/ionic-native/commit/7c13bae755076baa8cc50dc5f3d77156ba72a32b

wf9a5m75:
Oh, great!!!

wf9a5m75:
@ihadeed Sorry, could you tell me how to use the `wrapPromise`?
I tried it though, I could not success.

ihadeed:
@wf9a5m75 

That method is used by the `Cordova` decorator.

If we implement that code, you can do this:
```ts
@Cordova({
  destruct: true
})
someMethod(): Promise<any> { return; }
```

Then it will automatically pass arguments as an array instead of a single argument. 

We can change it so that it automatically detects if more than 1 argument was passed, and returns an array. This way we don't have to pass `destruct`, and we can have functions that would pass 1 argument, or an array of arguments. I think this would be the way to go, since the `on` method can return 1 or more args.

wf9a5m75:
Thank you for your lecture. I archived to pass the multiple values, passed from the JS side, to the event listener.
But I wonder one thing.

Since I prefer pure JS (it means kind of old user?), I prefer to receive these values like this:

```js
onMarkerClick(latLng: any, marker: any) {
  marker.showInfoWindow();
}
```

But it seems I saw this way:

```js
onMarkerClick(...parameters: any[]) {
  let latLng = parameters[0];
  let marker = parameters[1];
  marker.showInfoWindow();
}
```

Could you tell me which way is natural syntax for ionic users?

-----

The current wrapper plugin works like this,

```js
marker.on(GoogleMapsEvent.MARKER_CLICK).subscribe(this.onMarkerClick);

onMarkerClick(latLng) => {
  
}
```

I think `onMarkerClick(latLng: any, marker: any)` is better to support backward compatibility.

But if this way `onMarkerClick(...parameters: any[]) {}` is more natural for ionic users, I pass the values for this.

wf9a5m75:
By the way, could you tell me how to pass `...args` using decorator?

```js
export function CordovaInstance(opts: any = {}) {
  return (target: Object, methodName: string) => {
    console.log(target);
    return {
      value: function(...args: any[]) {
        console.log('args', args);
        return wrapInstance(this, methodName, opts).apply(this, args);
      },
      enumerable: true
    };
  };
}
```

wf9a5m75:
Okay, finally this code works

```js
export class MoveCameraZoomInPage {
  map: GoogleMap;

  constructor(public navCtrl: NavController, public navParams: NavParams, private googleMaps: GoogleMaps) {
  }

  ionViewDidLoad() {
    let self = this;
    setTimeout(function() {
      self.loadMap();
    }, 1000);
  }

  loadMap() {
    this.map = this.googleMaps.create('map_canvas');
    this.map.one(GoogleMapsEvent.MAP_READY).then(() => {
      let self = this;

      //self.map.on(GoogleMapsEvent.MAP_DRAG_START).subscribe(self.onMapDragStart);
      //self.map.on(GoogleMapsEvent.MAP_DRAG).subscribe(self.onMapDrag);
      //self.map.on(GoogleMapsEvent.MAP_DRAG_END).subscribe(self.onMapDragEnd);

      self.map.addMarker({
        position: {lat: 0, lng: 0},
        title: "hello world"
      }).then(self.onMarkerAdd.bind(self));

    });

    this.map.on(GoogleMapsEvent.MAP_CLICK).subscribe((latLng) => {
      let self = this;

      self.map.addMarker({
        position: latLng,
        title: latLng.toUrlValue()
      }).then(self.onMarkerAdd.bind(self));

    });
  }

  onMapDragStart() {
    console.log('onMapDragStart');
  }
  onMapDrag() {
    console.log('onMapDrag');
  }
  onMapDragEnd() {
    console.log('onMapDragEnd');
  }

  onMarkerAdd(marker:Marker) {
    marker.one(GoogleMapsEvent.MARKER_CLICK).then(this.onMarkerClick);
  }
  onMarkerClick(parameters: any[]) {
    let marker = parameters[1];

    marker.on(GoogleMapsEvent.INFO_CLICK).subscribe(() => {
      alert("info clicked");
    });
  }
}
```

In order to archive this, I changed the event names of the wrapper plugin, 
https://github.com/wf9a5m75/ionic-native/blob/73a21b1403a821983a87370dbbfb0f3a950a7ee6/src/%40ionic-native/plugins/google-maps/index.ts#L332-L360

and the JS plugin works with `Promise` and `FakeObservable`.
https://github.com/mapsplugin/cordova-plugin-googlemaps/blob/d128c1f4839ac316fcc1cac17560f6bf010ee66c/www/BaseClass.js

ihadeed:
Why did you create a `FakeObservable` class?

Wouldn't the following code work fine?
```ts
@Cordova({ observable: true })
on(eventName: string): Observable<any> { return; }
```

wf9a5m75:
Because `on` and others methods are not able to convert the instance from JS to the ionic.

The JS version did not have the promise and the subscribe  methods.

The biggest your code was this.

If the JS does not have the FakeObservable, the code becomes like this

```
marker1.on(GoogleMapsEvent.MARKER_CLICK, onMarkerClick);
marker2.on(GoogleMapsEvent.MARKER_CLICK, onMarkerClick);

onMarkerClick(...params: any[]) => {
  let latLng = params[0];
  let marker = params[1];

  // This does not work, because marker is instance of JS, not ionic wrapper's one.
  // marker.one(GoogleMapsEvent.INFO_CLICK).then(() => {
  //   alert("info clicked");
  // });

  // This works, but this leads many people confuse a lot.
  marker.one(GoogleMapsEvent.INFO_CLICK, () => {
    alert("info clicked");
  });
}
```

wf9a5m75:
https://github.com/mapsplugin/cordova-plugin-googlemaps/pull/1613#issuecomment-321100174

ihadeed:
@wf9a5m75 

I understand. But the `@Cordova` decorator can convert callbacks to `Promise` or `Observable`. We just have to omit the `sync: true` option.

wf9a5m75:
As far as I understand, `@Cordova` decorator is for static class, isn't it?
I think `@CordovaInstance` because BaseClass is not static class.

I think there probably some mistakes in the index.ts, especially property names, constructor parameters.

ihadeed:
@wf9a5m75 

`Cordova` and `CordovaInstance` work the same way. The only difference is:
- `Cordova` applies method calls to the plugin's `pluginRef`
- `CordovaIntance` applies method calls to `this._objectInstance`

wf9a5m75:
We haven't tested even 1/3 of all methods.

I appreciate if you join to the working channel on us, and join to the forked repo. Then you can modify, and we can also test it before merging to the ionic-team repo.

wf9a5m75:
Oh, I see.

ihadeed:
@wf9a5m75 

Great!

I'll clone the repos and check them out soon.

wf9a5m75:
Thank you for jumping on!
You don't need to work on the ionic-googlemaps-demo repo, but you can test it.

wf9a5m75:
@ihadeed Well, we haven't checked all features yet, but at least lots of problems are fixed.
I really appreciate if you merge the code at once.
Thanks.

(Asking the problems of the current @ionic-native/googlemaps has been increasing...)

wf9a5m75:
@ihadeed Could you update the npm at once?

I haven't checked all features, but the merge reduces troubles currently reported (user code is correct, but the current plugin is bad).

PR #1867, opened 2017-08-05
docs(navigation): add trailing slash to navigation items

janpio:
fixes the problem that each navigation item first has to redirected to the version with trailing slash

closes https://github.com/ionic-team/ionic-site/issues/1223


ihadeed:
@perrygovier would have to approve this change since it's site related.

perrygovier:
Fine by me

PR #1874, opened 2017-08-09
 Geofence plugin method not working maybe duplicate entry with one typo.

helderjnpinto:
I not sure but in Geofence plugin, when i call the method onTrasitionReceived() the ionic says the plugin is not installed, and it crash. exist other method with 'n' onTransitionReceived but it's working! in the doc's this is very confuse i spend more than a houre to discover this mess. I not sure why this method is there but if not return nothing maybe it's better removing this. Sorry for the english! :-)


ihadeed:
Just pushed a commit to fix this... `@CordovaFunctionOverride` is cleaner, the name of the method was just wrong.

PR #1900, opened 2017-08-19
Adding localizedReason in fingerprint-aio

AshfaqMemon:
I want to add localizedReason as optional string parameter in fingerprint-aio wrapper.


ihadeed:
This PR is too messy to review/merge.

PR #1906, opened 2017-08-21
Update index.ts

ryanhalley:
If you 'ionic cordova plugin add cordova-plugin-market' you get a 404; I'm guessing both plugin and pluginRef should be https://github.com/xmartlabs/cordova-plugin-market, as shown on the repo readme (which definitely works)


ihadeed:
Please update the PR title and description. "Update index.ts" doesn't explain what this PR is supposed to change.

ihadeed:
`pluginRef` is the reference to the plugin's object on `window`. Which is `cordova.plugins.market` in this case.

PR #1910, opened 2017-08-22
Update index.ts

yechenchao:
Adding GooglePlus to providers


ihadeed:
Please update the PR title and description. "Update index.ts" doesn't explain what this PR is supposed to change.

ihadeed:
The docs already explain this step.

PR #1911, opened 2017-08-22
Update index.ts

yechenchao:
Adding Facebook plugin to providers of component is required to make this plugin work.


ihadeed:

PR #1912, opened 2017-08-22
Update index.ts usage

yechenchao:
Adding splashScreen to providers is required to make this work


ihadeed:

PR #1913, opened 2017-08-22
Updating index.ts usage

yechenchao:
Adding statusBar to providers is needed


ihadeed:

PR #1917, opened 2017-08-23
add optional key parameter to SQLiteDatabaseConfig to support sqlcipher-adapter

rapropos:
As described in [this forum thread](https://forum.ionicframework.com/t/execution-failed-for-task-processdebugresources-ionic-compile/103001), users of [cordova-sqlitecipher-adapter](https://github.com/litehelpers/Cordova-sqlcipher-adapter) will encounter type errors when adding their needed encryption key to `SQLite.create()` calls. This PR adds it as an optional parameter.


ihadeed:
Thanks @rapropos

PR #1940, opened 2017-09-01
Fixed success & error index for launch-navigator.navigate

mflynnKVD:
Fixes an issue with the latest launch-navigator plugin (v4). Plugin expects options to be second parameter passed into method `navigate` but ionic-native wrapper was setting the success callback as the second input parameter.


rcrodrigues:
I believe this commit broke my app. It's actually opening the options for user to use when actually I`m setting it to open maps directly.  Reverting this solved my problem.

PR #1980, opened 2017-09-18
Added Extended Device Information

danielehrhardt:
Gets the System Memory and the CPU Information


src/@ionic-native/plugins/extended-device-information/index.ts
```diff
@@ -0,0 +1,40 @@
+import { Injectable } from '@angular/core';
+import { CordovaProperty, Plugin, IonicNativePlugin } from '@ionic-native/core';
+
+declare const window: any;
```
ihadeed: This is not needed.

danielehrhardt:
@ihadeed i removed this line

PR #1993, opened 2017-09-25
feat(cordova-plugin-certificates): add plugin

stovmascript:


ihadeed:
Could you please provide a use-case for this plugin?

stovmascript:
I'm currently developing an app for a partner who's dev/test server has a
self-signed SSL certificate. Their production environment has a standard
verified certificate. I need to be able to deploy to both environments,
without the hassle of digging in and editing the Cordova WebView client,
which isn't even checked into version control.

ihadeed:
May I suggest you use CloudFlare to proxy your server? This will allow you to keep your self-signed certificate, and cover it with a layer of a trusted SSL.

PR #2006, opened 2017-10-01
fix(indexappcontent): onItemPressed returns observable

NiklasMerz:
Now `onItemPressed` will return an Observable. This should improve adding an event.

Inspired by the 3dTouch plugin function `onHomeIconPressed`.

### DIFF
--- a/src/@ionic-native/plugins/index-app-content/index.ts
+++ b/src/@ionic-native/plugins/index-app-content/index.ts
@@ -1,5 +1,6 @@
 import { Injectable } from '@angular/core';
-import { Plugin, Cordova, IonicNativePlugin } from '@ionic-native/core';
+import { Plugin, Cordova, CordovaFunctionOverride, IonicNativePlugin } from '@ionic-native/core';
+import { Observable } from 'rxjs/Observable';
 
 export interface IndexItem {
   domain: string;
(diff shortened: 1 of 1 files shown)


ihadeed:
You can use the `@CordovaFunctionOverride` decorator. See example: https://github.com/ionic-team/ionic-native/blob/00c0707dad4fc98974ae2e8c3d720c7a7521e968/src/%40ionic-native/plugins/geofence/index.ts#L98

Sorry the decorator is not documented yet!

NiklasMerz:
I have changed this to the `@CordovaFunctionOverride()`.

NiklasMerz:
Is the the `@CordovaFunctionOverride` decorator working. I have some problems with it?

see also #2028 #2029

ihadeed:
@NiklasMerz it should be working, yes. I will test it soon and apply any fixes if necessary.

NiklasMerz:
@ihadeed I have some trouble with the latest change. I get warnings on the console that the onitemPressed function is not avaivalble. Do you know how the get the decorator working in this scenario?

ihadeed:
@NiklasMerz Found the issue, just pushed a commit to fix it. Update to the latest `@ionic-native/core` and it should be resolved. (Publishing it now, should be live in a few minutes)

NiklasMerz:
Thank you @ihadeed. This fixed the issue for me, too.

PR #2024, opened 2017-10-06
Fix: @ionic-native/google-maps 4.3.0 problems

wf9a5m75:
This PR solves these problems:

- GoogleMapsEvent.POLYGON_CLICK does not work

- The marker instance generated by MarkerCluster does not work
 (https://github.com/mapsplugin/cordova-plugin-googlemaps/issues/1706)

- The instances of wrapper class remains even the JS instances are removed.


wf9a5m75:
Thank you!

PR #2032, opened 2017-10-09
fix(firebase): optional params, callback indexes fix

stormwin:
Add `optional` label for optional parameters in docs
remove wrong callback indexes
make 'namespace' param optional in setDefaults


ihadeed:
The callback indexes are needed because we have an optional parameter. They are correct and do not need to be removed. See https://github.com/arnesson/cordova-plugin-firebase/blob/master/www/firebase.js#L115

Keep in mind indexes start from 0, not 1.

ihadeed:
You can edit the PR and only include the docs update, or close this.

stormwin:
Last time i tried this without this patch, it gave me very strange results both when i have and do not have a value for namespace...
Look at the code in plugin... it says if you do not provide namespace then error = success, success = namespace and that is index 1 and 2 and not 2 and 3 anymore
```
exports.setDefaults = function (defaults, namespace, success, error) {
    var args = [defaults];
    if (typeof namespace === 'string') {
        args.push(namespace);
    } else {
        error = success;
        success = namespace;
    }
    exec(success, error, "FirebasePlugin", "setDefaults", args);
};
```
I remember that in that case i got success response in error callback.
P.S: i already use this patch in apps that is in production, and they are working perfectly

ihadeed:
Oh I see. Sorry I didn't notice that earlier. I will take another look soon.

Regards,
Ibby Hadeed

danielsogl:
@stormwin Can you fix the conflicts?

PR #2041, opened 2017-10-12
feat(pro): support corova-plugin-ionic v2.0.0

Fuiste:
As of version 2.0.0 of `cordova-plugin-ionic`, the js API no longer requires app ID's passed with each call.  I've also refactored `init` to take a config object and fixed the `download` and `extract` finctions to properly use observables.


Fuiste:
Oops, looks like I forgot to lint.  I'll reopen with that fixed.

PR #2046, opened 2017-10-18
[docs] add missing `tapToFocus` camera option

VictorCoding:


src/@ionic-native/plugins/camera-preview/index.ts
```diff
@@ -36,6 +36,9 @@ export interface CameraPreviewOptions {
 
   /** Alpha channel of the preview box, float, [0,1], default 1 */
   alpha?: number;
+
+  /** Tap to set specific focus point. Note, this assumes the camera is full-screen. default false */
+  tapToFocus?: boolean;
```
janpio: Does `boolean` match the functionality of the original plugin?

PR #2060, opened 2017-10-24
Add senderID to AndroidPushOptions

cmorbitzer:
phonegap-plugin-push requires `senderID` as an Android option. This change allows TypeScript to acknowledge this parameter and enforce this requirement.

Type definition here: https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/phonegap-plugin-push/index.d.ts#L129

### DIFF
--- a/src/@ionic-native/plugins/push/index.ts
+++ b/src/@ionic-native/plugins/push/index.ts
@@ -132,6 +132,11 @@ export interface CategoryActionData {
 
 export interface AndroidPushOptions {
   /**
+   * Maps to the project number in the Google Developer Console.
+   */
+  senderID: string;
+
+  /**
    * The name of a drawable resource to use as the small-icon. The name should
    * not include the extension.
    */


trytuna:
Thats not true. 

> As of version 2.0.0 the SENDER_ID parameter has been removed at install time. Instead you put your google-services.json (Android) and/or GoogleService-Info.plist in the root folder of your project and then add the following lines into your config.xml.

You guys already did that correctly here https://github.com/ionic-team/ionic-native/commit/34bf136

🤦‍♂️

Todmy:
This changes broke my app :sob: 
We don't need to specify SENDER_ID, since it comes to the app from google-services.json file. And I think it is overhead to duplicate this key here.

trytuna:
@Todmy 

`private pushOptions: any;` should fix your problem temporarily.

Todmy:
@trytuna I know, thank you :+1:

PR #2061, opened 2017-10-24
[docs] Fix typo in class name @ionic-native/phonegap-local-notification

saoron:


janpio:
Are you sure this is correct? The interface is actually in plural, ending with an `s`: https://github.com/saoron/ionic-native/blob/master/src/%40ionic-native/plugins/phonegap-local-notification/index.ts#L24

Tests are also failing because of this:
```
src/@ionic-native/plugins/phonegap-local-notification/index.ts(13,39): error TS2304: Cannot find name 'LocalNotificationOption'.
src/@ionic-native/plugins/phonegap-local-notification/index.ts(13,39): error TS4063: Parameter 'options' of constructor from exported class has or is using private name 'LocalNotificationOption'.
```

I think you should close this PR.

Issue #1373, opened 2017-04-14
Feature Request - Braintree Plugin


wwallace:
Although it is possible to use the Braintree JS Sdk, it is somewhat limited and lacking certain features or is more difficult to integrate with a hybrid app, specifically the Braintree dropin ui

Dropin:
iOS: https://developers.braintreepayments.com/guides/drop-in/ios/v4
Android: https://developers.braintreepayments.com/guides/drop-in/android/v2

Cordova Plugin?
https://github.com/Taracque/cordova-plugin-braintree

borntodesign:
@wwallace could you provide a rough overview of how you currently have it integrated as I am looking into Braintree Marketplace.

danielsogl:
@wwallace I will add this plugin later this week

dcesca:
@danielsogl any news about the plugin for braintree?

codinronan:
Hey guys,

I have put this together and opened a PR:
https://github.com/ionic-team/ionic-native/pull/2073

Hopefully we can get this in. Works great locally.

PR #2073, opened 2017-10-29
feat(braintree): add plugin support

codinronan:
## Description

Adds support for the Braintree Drop-In Payments UI via `cordova-plugin-braintree`.

NOTE: Does not use the npm version of `cordova-plugin-braintree`, which is out of date and not maintained. This is intended to be used by a maintained fork at https://github.com/Taracque/cordova-plugin-braintree

Addresses #1373 

@danielsogl 
@wwallace


codinronan:
Hey guys, is there a regular review cycle for these or something? Don't mind waiting, just like to know what to expect.

janpio:
From what I know (and have seen) there is not. Merges happen when they happen, sometimes a lot, sometimes not so much. Seems to be the latter right now.

PR #2082, opened 2017-11-02
Cordova plugin country

JahlomP:


ihadeed:
Isn't this the same thing as the `getLocaleName` method in the `Globalization` plugin?

PR #2085, opened 2017-11-02
support startStateNotifications & stopStateNotifications

mburger81:


mburger81:
@ihadeed Can you please merge this PR, it is a small change with two new functions

src/@ionic-native/plugins/ble/index.ts
```diff
@@ -417,6 +417,33 @@ export class BLE extends IonicNativePlugin {
   @Cordova()
   isEnabled(): Promise<void> { return; }
 
+  /**
+   * Register to be notified when Bluetooth state changes on the device.
+   *
+   * @usage
```
ihadeed: Please remove the usage block from here. This will not be displayed in the actual docs. You should document the method in the JSDoc of the class.

ihadeed: Never mind, the whole file is not formatted properly. I will fix it later.

lulumahot:
Hello i can't find this update in the latest release installed from npm (4.52). Is the npm module up to date with this ?

PR #2086, opened 2017-11-02
feat(DocumentPicker) adds DocumentPicker ios plugin

entomb:


ihadeed:
I think this plugin should be renamed to `IOSDocumentPicker` to reflect that it's an `iOS` specific plugin.

PR #2050, opened 2017-10-18
[add] Added support for iOS in @ionic-native/nfc

lehrenberg:

PR #2088, opened 2017-11-03
feat(NFC): add support for iOS

niconaso:
The last version of NFC plugin now supports NFC reading on iOS. I added the two required methods: beginSession and invalidateSession that need to be executed before and after reading.


lehrenberg:
Hi @niconaso I actually did that already a few weeks ago.

PR #2050

niconaso:
@lehrenberg excellent! I'll close this PR. Thanks!

Issue #2087, opened 2017-11-02
map.addMarker(MarkerOptions) restrict own property


wf9a5m75:
**I'm submitting a ...**  (check one with "x")
[x] bug report
[ ] feature request

**Current behavior:**

The cordova-plugin-googlemaps allows to users to include own property, but `@ionic-native/google-maps` refuse this.

This is original JS version code
```js
map.addMarker({
  position: {lat: ..., lng: ...},
  myData: {
    "hello": "world"
  }
}, function(marker) {
  console.log(marker.get("myData").get("hello"));
});
```

**Expected behavior:**

The `@ionic-native/google-maps` should work the same way.
```js
map.addMarker({
  position: {lat: ..., lng: ...},
  myData: {
    "hello": "world"
  }
}).then((marker: Marker) => {
  console.log(marker.get("myData").get("hello"));
});
```

wf9a5m75:
Oh, my mistake. It's work actually with the `@ionic-native/google-maps@4.3.3`. Forget about this.

wf9a5m75:
![screen shot 2017-11-06 at 1 00 31 pm](https://user-images.githubusercontent.com/167831/32463752-81db2e58-c2f2-11e7-9900-170397656979.png)

```
  "dependencies": {
    "@angular/common": "4.4.4",
    "@angular/compiler": "4.4.4",
    "@angular/compiler-cli": "4.4.4",
    "@angular/core": "4.4.4",
    "@angular/forms": "4.4.4",
    "@angular/http": "4.4.4",
    "@angular/platform-browser": "4.4.4",
    "@angular/platform-browser-dynamic": "4.4.4",
    "@ionic-native/core": "^4.3.3",
    "@ionic-native/google-maps": "^4.3.3",
    "@ionic-native/splash-screen": "4.3.2",
    "@ionic-native/status-bar": "4.3.2",
    "@ionic/storage": "2.0.1",
    "cordova-android": "6.3.0",
    "cordova-plugin-device": "^1.1.6",
    "cordova-plugin-googlemaps": "^2.1.1",
    "cordova-plugin-ionic-webview": "^1.1.16",
    "cordova-plugin-splashscreen": "^4.0.3",
    "cordova-plugin-whitelist": "^1.3.2",
    "ionic-angular": "3.8.0",
    "ionic-plugin-keyboard": "^2.2.1",
    "ionicons": "3.0.0",
    "rxjs": "5.4.3",
[...]

PR #2093, opened 2017-11-06
Add: accept own properties for `addMarker()` and others [google-maps]

wf9a5m75:
Add: accept own properties for `addMarker()` and others #2087

Fix: some properties are required, but specified as optional.

Add: descriptions for properties


wf9a5m75:
Thank you for merging!

PR #2094, opened 2017-11-07
fix (music-controls) make optional items optional, add hasScrubbing value for lock screen seeking

ghenry22:
Add hasScrubbing variable to options - used to enable or disable scrubbing from the iOS control center or lockscreen progress bar

Make non required options actually non-required.

Correct errors in usage examples.

### DIFF
--- a/src/@ionic-native/plugins/music-controls/index.ts
+++ b/src/@ionic-native/plugins/music-controls/index.ts
@@ -3,22 +3,23 @@ import { Plugin, Cordova, IonicNativePlugin } from '@ionic-native/core';
 import { Observable } from 'rxjs/Observable';
 
 export interface MusicControlsOptions {
-  track: string;
-  artist: string;
-  cover: string;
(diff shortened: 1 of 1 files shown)


ghenry22:
@ihadeed - would be great if you could merge this one, only small changes, all working in my own apps for the last few days of testing.

For change log:
fix(music-controls) - optional parameters are OPTIONAL
fix(music-controls documentation) - fix some errors in the usage examples
new(music-controls hasScrubbing option) - optional variable to enable lock screen seeking with progress bar on iOS.

ghenry22:
actually I have just matched them to the plugin itself.  So if it is optional for the plugin then it should be optional for the wrapper.

The plugin will apply default values to meet the basic requirements if you provide no values.

I agree that using the plugin with no options is kind of pointless but this PR just matches the bionic native wrapper up with the plugin itself.

joshbabb:
I should have looked closer.  Kudos, looks great!

PR #2098, opened 2017-11-07
NFC: add interfaces for NdefEvent, NdefRecord, NdefTag.

twaldecker:
Add interfaces for the result of addNdefListener


twaldecker:
There are coming more fixes. I am closing the PR for now.

PR #2145, opened 2017-11-24
In SocialSharing plugin, the provider is added

egesu:
The provider should be added in module declaration or we get `No provider for SocialSharing!` error.


egesu:
Ah sorry, I didn't notice it.

PR #2151, opened 2017-11-27
feat(swt-x9): Add support for the Shenzen Wearable Technologies X9 smart wristband

joshbabb:
See https://github.com/BlyncSync/cordova-plugin-swt-x9 for the underlying Cordova plugin (or just `npm install cordova-plugin-swt-x9`)


joshbabb:
@ihadeed Hey, is this a feature that won't be merged and should be integrated by us some other way?

joshbabb:
Well, if it isn't, a convenient way of designating that an app should use an entirely-different repository for all things Ionic Native would work!

PR #2180, opened 2017-12-06
feat(android exoplayer): add cordova-plugin-exoplayer

mebibou:
plugin: https://github.com/frontyard/cordova-plugin-exoplayer


syedhope:
Is there a way to play audio only?

The param {'audioOnly': false} throws an error => not assignable to parameter of type 'AndroidExoPlayerParams'.
Types of property 'audioOnly' are incompatible.
Type 'string' is not assignable to type 'true'.

PR #2190, opened 2017-12-10
Update ionic native - HTTP

moathdev:
Dears , 

Many developers complain about the plugin not working and the answer is easy 
just added to app.module.ts :)

### DIFF
--- a/src/@ionic-native/plugins/http/index.ts
+++ b/src/@ionic-native/plugins/http/index.ts
@@ -54,6 +54,7 @@ export interface HTTPResponse {
  *   });
  *
  * ```
+ * Note : Do not forget to add HTTP in the app.module.ts .
  * @interfaces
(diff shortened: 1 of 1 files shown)


raghav-khunger:
Yes, that comment will be helpful.

danielsogl:
Sorry but this is not necessary. Read the docs and you won't run into that issue: https://ionicframework.com/docs/native/

Checkout the usage section.

> Add Plugins to Your App's Module
> 
> After installing a plugin’s package, add it to your app’s NgModule.

moathdev:
its good and helpful , but can we put it somewhere visible for the new users ?

ihadeed:
It doesn't make sense to include this type of documentation to each plugin manually. The current docs template already includes instructions similar to this.

PR #2250, opened 2018-01-04
fix(google-analytics): add getVar and setVar functions

SpellChucker:
https://github.com/danwilson/google-analytics-plugin/blob/master/www/analytics.js#L36-L42


danielsogl:
I already added these two functions. Thank you.

PR #2257, opened 2018-01-09
fix(web-intent): rename `onNewIntent` to `onIntent`

FdezRomero:
The `onNewIntent` function doesn't exist in `darryncampbell-cordova-plugin-intent`, it's actually called `onIntent`.

Source: https://github.com/darryncampbell/darryncampbell-cordova-plugin-intent/blob/master/www/IntentShim.js#L62

This PR renames `onNewIntent` to `onIntent`, as the implementation was right (should return an Observable). It also removes the previous `onIntent` as it was wrong and shouldn't return `void` anyway.

The change has been tested on device with latest `ionic-angular@3.9.2` and over `@ionic-native/web-intent@4.5.2`.

### DIFF
--- a/src/@ionic-native/plugins/web-intent/index.ts
+++ b/src/@ionic-native/plugins/web-intent/index.ts
@@ -147,12 +147,13 @@ export class WebIntent extends IonicNativePlugin {
   getUri(): Promise<string> { return; };
 
   /**
+   * Returns the content of the intent used whenever the application activity is launched
    * @returns {Observable<string>}
    */
   @Cordova({
     observable: true
   })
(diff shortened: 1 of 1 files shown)


FdezRomero:
@danielsogl I forgot to mention that it would be nice to get this ported to 5.0.0, as it has the same issue.

FdezRomero:
Hi there! This PR has been approved, can we get it merged into master for the next release? Thanks!

danielsogl:
Can you resolve the conflicts?

FdezRomero:
@danielsogl I resolved the conflicts, they were because of the new linter rules 😉

PR #2268, opened 2018-01-15
feat(plugin): add Sensors plugin

Simpler1:
One plugin.  Many sensors.


danielsogl:
@Simpler1 can you add an TYPE_SENSOR enum to the plugin? Then I will merge it.

Simpler1:
@danielsogl I added an enum for TYPE_SENSOR.  Let me know if you need any other changes.

I also squashed to a single commit.

Thanks

PR #2276, opened 2018-01-17
feat(plugin): add cordova-hot-code-push

mebibou:
This is different from the current `code-push` which is tied to the Microsoft API, this plugin does not require to use Microsoft and can work with any server.

One thing I didn't add in the plugin: https://github.com/nordnet/cordova-hot-code-push/wiki/Listen-for-update-events. I am not sure if I can map these events somehow in the plugin


ihadeed:
You have two options for events:
```ts
@Cordova({
  eventObservable: true,
  event: 'chcp_updateIsReadyToInstall'
})
onUpdateIsReadyToInstall(): Observable<any> { return; }
```

Or you can do a manual function `on(eventName: string)` that returns an observable & subscribes to the event.

mebibou:
@ihadeed thanks, I've added the events

PR #2278, opened 2018-01-18
fix(cordova-file-plugin): write in an existing file with ArrayBuffer

fishme:
Current implantation for the feature `writeExistingFile` don't allow ArrayBuffer.


fishme:
I have no idea from where the conflicts are. I did also merge the master into my branch and there were no conflicts. Any Ideas?

danielsogl:
Added wit another PR.

Issue #2285, opened 2018-01-23
'postTemplate' does not exist in type 'BackgroundGeolocationConfig'


agenterp:
**I'm submitting a ...**  (check one with "x")
[x] bug report
[ ] feature request
[ ] support request => Please do not submit support requests here, use one of these channels: https://forum.ionicframework.com/ or https://ionicworldwide.herokuapp.com/

**Current behavior:**
The Plugin Configuration Parameter 'postTemplate' could not be set with the background-geolocation plugin.

**Expected behavior:**
Set the  'postTemplate' Configuration Parameter as described in the https://github.com/mauron85/cordova-plugin-background-geolocation Plugin

**Steps to reproduce:**
Add 
      postTemplate: {
        provider: '@provider',
        latitude: '@latitude',
        longitude: '@longitude',
        time: '@time',
        accuracy: '@accuracy',
        speed: '@speed',
        altitude: '@altitude',
        bearing: '@bearing',
        foo: 'bar'
      }
to the BackgroundGeolocationConfig.

``

**Ionic info:** (run `ionic info` from a terminal/cmd prompt and paste output below):

```
insert the output from ionic info here
```
cli packages: (/usr/lib/node_modules)

    @ionic/cli-utils  : 1.19.1
    ionic (Ionic CLI) : 3.19.1

global packages:

    cordova (Cordova CLI) : 8.0.0 

local packages:

    @ionic/app-scripts : 3.1.2
    Cordova Platforms  : android 7.0.0 browser 5.0.3
    Ionic Framework    : ionic-angular 3.9.2

System:

    Android SDK Tools : 26.1.1
    Node              : v6.12.2
    npm               : 3.10.10 
    OS                : Linux 4.8

Environment Variables:

    ANDROID_HOME : /home/deu/Android/Sdk

Misc:

    backend : pro

agenterp:
Added postTemplate and send a Pull Request https://github.com/ionic-team/ionic-native/pull/2286

PR #2286, opened 2018-01-23
Added postTemplate to BackgroundGeolocationConfig Issue #2285

agenterp:


danielsogl:
I close this PR because of the upcoming v3 implementation

saborpl:
@danielcregg  v3 implementation of ionic-native? i can't find this...

malrok44:
I think @danielsogl was talking about the v3 of cordova-plugin-mauron85-background-geolocation.
As they mention on their github page, the documentation is for v3 which is still in alpha.
And the postTemplate option is found only in v3.

Faizulla-new:
in backgroundGeoLocation postTemplate is not working

PR #2292, opened 2018-01-26
docs(pro): add types for return values in deploy

Maistho:
Adding some better types on the ProDeploy return values

### DIFF
--- a/src/@ionic-native/plugins/pro/index.ts
+++ b/src/@ionic-native/plugins/pro/index.ts
@@ -31,6 +31,8 @@ export interface DeployConfig {
   channel?: string;
 }
 
+export type ProgressMessage = number | string;
+
 /**
  * @hidden
  */
@@ -43,7 +45,7 @@ export class ProDeploy {
    * @param config A valid Deploy config object
    */
   @CordovaInstance()
-  init(config: DeployConfig): Promise<any> { return; }
+  init(config: DeployConfig): Promise<void> { return; }
 
   /**
    * Check a channel for an available update
@@ -54,27 +56,27 @@ export class ProDeploy {
 
   /**
    * Download an available version
-   * @return {Observable<any>} Updates with percent completion, or errors with a message.
+   * @return {Observable<ProgressMessage>} Updates with percent completion, or errors with a message.
    */
   @CordovaInstance({
     observable: true
   })
-  download(): Observable<any> { return; }
+  download(): Observable<ProgressMessage> { return; }
 
   /**
    * Unzip the latest downloaded version
-   * @return {Observable<any>} Updates with percent completion, or errors with a message.
+   * @return {Observable<ProgressMessage>} Updates with percent completion, or errors with a message.
    */
   @CordovaInstance({
     observable: true
   })
-  extract(): Observable<any> { return; }
+  extract(): Observable<ProgressMessage> { return; }
 
   /**
    * Reload app with the deployed version
    */
   @CordovaInstance()
-  redirect(): Promise<any> { return; }
+  redirect(): Promise<void> { return; }
 
   /**
    * Get info about the version running on the device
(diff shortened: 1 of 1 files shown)


ihadeed:
`number | string` can work when you have an interface defining a parameter (write-only usage) .. it causes issues when having a `read/write` or `read-only` interface. 

Example:
If you try to do `value.toUpperCase()` .. the IDE/Compiler will say that `toUpperCase` does not exist on type `number`.

Maistho:
But that's the point of type checking? If I try to use `toUpperCase()` on a number it would be good if the compiler complains so that I can fix my code?

ihadeed:
Here's another example of what I mean,  hopefully it will make sense:
```ts
download()
  .subscribe((value: ProgressMessage) => {
    // TypeScript isn't sure whether `value` is a string or a number unless if we explicitly set the type
    // if we explicitly set the type, I don't see a point of having a type in the first place. 
  });
```

Maistho:
Yes typescript isn't sure whether `value` is a string or a number, and usually we (programmers) are not sure either. Usually the code that runs `download` would have to check what the return value is.

See the following example for what I would think is a common use case:

```typescript
this.pro.download().subscribe((value: ProgressMessage) => {
  if (typeof value === 'string') {
    // We are either done or it failed we probably would want to check for 'done' and
    // if so move on with extracting, or else show an error message to the user

    // Here typescript will detect that value is a string within this block, so
    // `value.toUpperCase()` for example will work
    console.log(value.toUpperCase()); // This works!
  } else {
    // This is a progress update, we probably want to update a progress bar or similar

    // Typescript will detect that value is a number within this block, so 
    // `value.toExponential()` for example will work.
    console.log(value.toExponential()); // This works!
  }
});
```

ihadeed:
I think you need to do `(value as string).toUpperCase()` or `(value as number).toExponential()` to avoid TS issues.

Maistho:
Nope, TS is smart enough to [infer](https://www.typescriptlang.org/docs/handbook/type-inference.html) the correct types even with [union types](https://www.typescriptlang.org/docs/handbook/advanced-types.html#union-types).

> Luckily, you don’t need to abstract `typeof x === "number"` into its own function because TypeScript will recognize it as a type guard on its own. 

See [this example](https://www.typescriptlang.org/play/#src=type%20ProgressMessage%20%3D%20number%20%7C%20string%3B%0D%0A%0D%0Aconst%20handler%20%3D%20(value%3A%20ProgressMessage)%20%3D%3E%20%7B%0D%0A%20%20%20%20if%20(typeof%20value%20%3D%3D%3D%20'string')%20%7B%0D%0A%20%20%20%20%20%20%20%20%2F%2F%20We%20are%20either%20done%20or%20it%20failed%20we%20probably%20would%20want%20to%20check%20for%20'done'%20and%0D%0A%20%20%20%20%20%20%20%20%2F%2F%20if%20so%20move%20on%20with%20extracting%2C%20or%20else%20show%20an%20error%20message%20to%20the%20user%0D%0A%0D%0A%20%20%20%20%20%20%20%20%2F%2F%20Here%20typescript%20will%20detect%20that%20value%20is%20a%20string%20within%20this%20block%2C%20so%0D%0A%20%20%20%20%20%20%20%20%2F%2F%20%60value.toUpperCase()%60%20for%20example%20will%20work%0D%0A%20%20%20%20%20%20%20%20console.log(value.toUpperCase())%3B%20%2F%2F%20This%20works!%0D%0A%20%20%20%20%7D%20else%20%7B%0D%0A%20%20%20%20%20%20%20%20%2F%2F%20This%20is%20a%20progress%20update%2C%20we%20probably%20want%20to%20update%20a%20progress%20bar%20or%20similar%0D%0A%0D%0A%20%20%20%20%20%20%20%20%2F%2F%20Typescript%20will%20detect%20that%20value%20is%20a%20number%20within%20this%20block%2C%20so%20%0D%0A%20%20%20%20%20%20%20%20%2F%2F%20%60value.toExponential()%60%20for%20example%20will%20work.%0D%0A%20%20%20%20%20%20%20%20console.log(value.toExponential())%3B%20%2F%2F%20This%20works!%0D%0A%20%20%20%20%7D%0D%0A%7D%3B%0D%0A) at the typescript playground

Maistho:
Any progress on this PR @ihadeed ?

Maistho:
Hey, I rebased it since there was 333 commits behind. This should be an easy merge?

PR #2294, opened 2018-01-28
feat(call-log): add plugin

VoltX:
Hello,
here is a wrapper of our cordova-plugin-calllog, a plugin to get the call log of device.
Thanks !


VoltX:
@ihadeed can you tell me how to proceed if i corrections need to be made to this pull-request? Should I cancel and start a new one? Thank you

PR #2316, opened 2018-02-02
feat(firebase-dynamic-links): update to conform to the latest version of the plugin

adamduren:
*Summary*

The `sendInvitation` method was removed as well as the related configuration variables. Also `onDynamicLink ` can be called multiple times during the app lifecycle so it makes more sense that it is an Observable.


adamduren:
@danielsogl thanks for the approval. How's responsible for merging?

bhavinwebdesk:
@danielsogl can you please merge it?

danielsogl:
cc @ihadeed

bhavinwebdesk:
@adamduren Just to be sure. This pull request has to change index.d.ts instead of index.ts if I am not wrong.

adamduren:
@bhavinwebdesk That file in generated during the build step and is not committed if I understand correctly.

PR #2320, opened 2018-02-07
Update for the native-google-maps plugin

thezakman:


wf9a5m75:
We have been waiting for the next updating the `@ionic-native/google-maps` so far.

PR #2332, opened 2018-02-13
docs(camera-preview): Added critical note to camera preview native plugin

khevamann:
The toBack mode does not work without adding this css style. This code sets the background of all ionic elements to transparent so that the camera preview will show up behind the elements

### DIFF
--- a/src/@ionic-native/plugins/camera-preview/index.ts
+++ b/src/@ionic-native/plugins/camera-preview/index.ts
@@ -121,7 +121,9 @@ export interface CameraPreviewPictureOptions {
  * this.cameraPreview.stopCamera();
  *
  * ```
- *
+ * @Important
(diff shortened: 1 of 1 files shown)


danielsogl:
Hey @khevamann, can you please describe this notice a bit more as @nuno mentioned?

khevamann:
@danielsogl I forgot to submit my reply to @nuno check above

danielsogl:
But then you should write that note like "add this css value to your page: background-color: transparent !important;". Because your given class is not generic.

khevamann:
@danielsogl  I accidentally already deleted this repository from my account so I can't edit the code. What should I do? Close this one and re open a new one?

danielsogl:
Yes reopen it

PR #2348, opened 2018-02-24
docs(onesignal): change docs to match new Cordova hook specs

karrirasinmaki:
- Script to match module definition (fixes paths)  
- Instructions to define hooks in recommend way


karrirasinmaki:
Sorry guys, forget to read thru and follow your commit message guideline before committing.

PR #2351, opened 2018-02-28
fix(crop): fix targetHeight, tagetWidth

akshaydwivedi29:
fix crop targetHeight, tagetWidth as optional parameters


danielsogl:
I implemented this by adding an options interface to this plugin.

PR #2353, opened 2018-03-01
feat(music-controls): add new optional parameters for icons

pniraula:
This will add optional parameters for icons in android.


pniraula:
Why does it take so long for completing pull requests? Am I missing something or it's just getting ignored?

PR #2406, opened 2018-03-17
feat(plugin): add OpenALPR plugin

iMicknl:
Adds support for [cordova-plugin-openalpr](http://github.com/iMicknl/cordova-plugin-openalpr). Just tested this with the v2.0 version of our plugin and it works like a charm.

Will all documentation be generated from the index.ts or is there a place where I need to commit this?


iMicknl:
@danielsogl, I see you add the 'needs testing' label. If you want to test it yourself, you need to use the develop branch for now (https://github.com/iMicknl/cordova-plugin-openalpr/tree/develop). If necessary I could also publish it on NPM as a @preview build. 

We are working on some last tests regarding the scanning engine itself. The Typescript / Ionic Native bindings are working great. :-)

PR #2412, opened 2018-03-21
Update this.sms.send() example usage

frizzkitten:
This function only works if you have an n before the number, e.g. `n1112223333`.

### DIFF
--- a/src/@ionic-native/plugins/sms/index.ts
+++ b/src/@ionic-native/plugins/sms/index.ts
@@ -41,8 +41,8 @@ export interface SmsOptionsAndroid {
  * ...
  *
  *
- * // Send a text message using default options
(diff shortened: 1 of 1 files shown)


danielsogl:
Hey @frizzkitten,
I checked the official plugin docs and can't find an instruction to use this plugin like that.

frizzkitten:
I don't know why the official plugin docs don't say to use it that way, because I tried using it the way they recommend and it didn't work. After hours of debugging, I randomly found that the only way to make a text send was by putting an 'n' before the number. Without it, apps would crash on calling the function. If you want to see the same results, try cloning and using this [example repo](https://github.com/PaulHalliday/Ionic2_SMS). This was on motox4 Android 8.0.0.

PR #2441, opened 2018-04-07
feat(plugin): Add google nearby plugin

hahahannes:
Hi, I would like to merge a new plugin that wraps [this cordova plugin ](https://github.com/hahahannes/cordova-plugin-google-nearby) with support for the Google Nearby API. At the moment it only supports android and the Messages API not the Connections API.

Thanks!
Hannes

### DIFF
--- a/src/@ionic-native/plugins/google-nearby/index.ts
+++ b/src/@ionic-native/plugins/google-nearby/index.ts
@@ -27,7 +27,7 @@ import { Observable } from 'rxjs/Observable';
   pluginName: 'GoogleNearby',
   plugin: 'cordova-plugin-google-nearby',
   pluginRef: 'window.nearby',
-  repo: 'https://github.com/hahahannes/googlenearby-cordova-plugin',
+  repo: 'https://github.com/hahahannes/cordova-plugin-google-nearby',
   install: 'ionic cordova plugin add cordova-plugin-google-nearby --variable API_KEY="123456789"',
   installVariables: ['API_KEY'],
   platforms: ['Android']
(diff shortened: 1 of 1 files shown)


src/@ionic-native/plugins/google-nearby/index.ts
```diff
@@ -0,0 +1,69 @@
...
+ * this.googleNearby.subscribe()
+ *   .then((res: any) => console.log(res))
+ *   .catch((error: any) => console.error(error));
+ * ```
+ */
+@Plugin({
+  pluginName: 'GoogleNearby',
+  plugin: 'cordova-plugin-google-nearby',
+  pluginRef: 'window.nearby',
+  repo: 'https://github.com/hahahannes/googlenearby-cordova-plugin',
```
ihadeed: Repo url should be `https://github.com/hahahannes/cordova-plugin-google-nearby`

src/@ionic-native/plugins/google-nearby/index.ts
```diff
@@ -0,0 +1,69 @@
...
+  subscribe(): Observable<any> {
+    return;
+  }
+
+/**
+   * Unsubscribe from Messages
+   * @return {Promise<any>} Returns a promise that resolves when you unsubscribed
+   */
+  @Cordova()
+  unsubscribe(): Promise<any> {
```
ihadeed: This method is not needed

hahahannes: Okay, I understand because it is enough to call Observable.unsubscribe() ?

danielsogl: Yes :)

PR #2450, opened 2018-04-12
fix(web-intent): fix options param

mcelotti:
fix(web-intent): fix options param for startService, sendBroadcast, startActivityForResult, startActivity


mcelotti:
Hi @danielsogl I've updated the code, thank you.

PR #2463, opened 2018-04-17
docs(File): Add template literals to escape punctuation

nioperas06:


src/@ionic-native/plugins/file/index.ts
```diff
@@ -271,7 +271,8 @@ export interface DirectoryEntry extends Entry {
 
   /**
    * Creates or looks up a file.
-   * @param path Either an absolute path or a relative path from this DirectoryEntry to the file to be looked up or created. It is an error to attempt to create a file whose immediate parent does not yet exist.
+   * @param path Either an absolute path or a relative path from this DirectoryEntry to the file to be looked up or created. It is an error to attempt to create a file whose immediate parent 
+   not yet exist.
```
danielsogl: Broken JSDOC

src/@ionic-native/plugins/file/index.ts
```diff
@@ -654,7 +655,7 @@ declare const window: Window;
  *
  * ...
  *
- * this.file.checkDir(this.file.dataDirectory, 'mydir').then(_ => console.log('Directory exists')).catch(err => console.log('Directory doesn't exist'));
+ * this.file.checkDir(this.file.dataDirectory, 'mydir').then(_ => console.log('Directory exists')).catch(err => console.log(`Directory doesn't exist`));
```
danielsogl: why is this necessary because an plain string is printed

PR #2503, opened 2018-05-10
stop transforming "declare" classes in v5

nicolashenry:
This PR fixes a wrong behaviour in v5 which transform "declare" classes to real classes.

For example the ```file``` plugin declare a class like this : 
```ts
export declare class FileReader {
   ...
}
```
and currently in the final ```file``` plugin package, v5 turns it into this : 
```js
var FileReader = /** @class */ (function () {
    function FileReader() {
    }
    return FileReader;
}());
export { FileReader };
``` 
which is wrong because FileReader should come from ```cordova-plugin-file```.

### DIFF
--- a/scripts/build/transformers/plugin-class.ts
+++ b/scripts/build/transformers/plugin-class.ts
@@ -57,7 +57,8 @@ function transformClasses(
   return ts.visitEachChild(
     file,
     node => {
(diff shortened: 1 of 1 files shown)


ihadeed:
Good catch, thanks!

PR #2505, opened 2018-05-14
fix(native-geocoder) result is array of objects

ciekawy:
as described in current version - https://github.com/sebastianbaar/cordova-plugin-nativegeocoder/ - promises resolve with an array

also confirmed this on iOS real device


ihadeed:
thanks

PR #2506, opened 2018-05-15
feat(local-notifications): added support for foreground option in ver…

mino922:
…sion 0.9.0-beta.3 of cordova-plugin-local-notifications

The foreground option is not support and launch is not working as expected with that latest version of the plugin.

### DIFF
--- a/src/@ionic-native/plugins/local-notifications/index.ts
+++ b/src/@ionic-native/plugins/local-notifications/index.ts
@@ -125,7 +125,7 @@ export interface ILocalNotificationAction {
   /**
    * Specifies whether the action causes the app to launch in the foreground
    */
-  launch?: boolean;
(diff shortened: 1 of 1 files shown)


src/@ionic-native/plugins/local-notifications/index.ts
```diff
@@ -125,7 +125,7 @@ export interface ILocalNotificationAction {
   /**
    * Specifies whether the action causes the app to launch in the foreground
    */
-  launch?: boolean;
```
ihadeed: We should keep `launch` to maintain backwards compatibility. No need to introduce a breaking change yet.

mino922: I agree. I'll go ahead and update that.

PR #2507, opened 2018-05-15
fix(OSNotification) can't create notifications

khevamann:
Made all properties in OSNotification optional because when you create a push notification the type is OSNotification but it does not have all the same properties required as the OSNotification as a response from a received notification. For example, you don't pass in isAppInFocus when you create a notification but it was causing issues when trying to post a notification because it was showing as required because OSNotification as a response will always have isAppInFocus.


ihadeed:
thanks

PR #2508, opened 2018-05-16
fix(android-fingerprint-auth): ensure ERRORS array is defined

msalcala11:
Currently, despite being documented here (https://ionicframework.com/docs/native/android-fingerprint-auth/#ERRORS), the `androidFingerprintAuth.ERRORS` array is undefined at run time. This PR ensures the array is available at run time.


ihadeed:
thanks!

Issue #2521, opened 2018-05-28
ThemeableBrowser cannot hidden and clear cache


hiepxanh:
**I'm submitting a ...**  (check one with "x")
[x] bug report
[ ] feature request
[ ] support request => Please do not submit support requests here, use one of these channels: https://forum.ionicframework.com/ or https://ionicworldwide.herokuapp.com/

**Current behavior:**

`this.themeableBrowser.create(this.endpoint, '_blank', {hidden:true});`

```
Argument of type '{ hidden: boolean; }' is not assignable to parameter of type 'ThemeableBrowserOptions'. 
Types of property 'hidden' are incompatible. Type 'boolean' is not assignable to type 'string'. 
```

**Expected behavior:**
`this.themeableBrowser.create(this.endpoint, '_blank', {hidden:true});
`
**Steps to reproduce:**

**Related code:**

```
insert any relevant code here
```

**Other information:**

**Ionic info:** (run `ionic info` from a terminal/cmd prompt and paste output below):

```
insert the output from ionic info here
```

hiepxanh:
waiting for an fixing [PR](https://github.com/ionic-team/ionic-native/issues/2521)

PR #2522, opened 2018-05-28
fix(themeable-browser): allow hidden and clear cache

hiepxanh:
it wrong from boolean to string, after 1 year i update it. already test many time.

clear for issue: [ThemeableBrowser cannot hidden and clear cache](https://github.com/ionic-team/ionic-native/issues/2521)

closes: #2521


hiepxanh:
can't believe no one fix it after almost 1 year...

PR #2541, opened 2018-06-08
Add cordova-plugin-ionic-webview with convertFileSrc function

jcesarmobile:
Related
https://github.com/ionic-team/ionic-cli/issues/3279

@dwieeb

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -60,6 +60,16 @@
       "integrity": "sha512-El3+WJk2D/ppWNd2X05aiP5l2k4EwF7KwheknQZls+I26eSICoWRhRIJ56jGgw2dqNGQ5LtNajmBU2ajS28EvQ==",
       "dev": true
     },
+    "JSONStream": {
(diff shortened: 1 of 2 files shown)


dwieeb:
Does it make sense for the URL conversion logic to be in Ionic Native or the Webview Plugin itself? I think the design concept of Ionic Native is to be more of a modern API wrapper around plugin code.

dwieeb:
I think the only purpose to have it here is it's just the "angular" way. People can test it easier, it's more familiar.

Otherwise they'd have to do something like this:

```ts
declare var window: any;
...
window.convertFileSrc()
```

jcesarmobile:
Partly agree with @danielsogl, but that function will be required to make next version of webview plugin to properly work, so we can't tell users they have to add their own convertFileSrc function, we have to provide it

dwieeb:
👉 https://github.com/ionic-team/ionic-native/pull/2549

PR #2544, opened 2018-06-09
fix(health-kit): add missing checkAuthStatus param

CosmicWebServices:
checkAuthStatus requires the input type but it is not available in healthKitOptions


danielsogl:
Please fix the build errors

PR #2547, opened 2018-06-12
fix(background-geolocation): watchLocationMode returns observable

frankkoenigstein:
# Background Geolocation `watchLocationMode` Observable
Plugin has a method to watch location mode: `watchLocationMode`.
Currently it is declared as returning a Promise. That means the callback is only called ones.

Since this is a listener that is called multiple times from the native part of the plugin it has to be declared to return an observable that can be subscribed.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "ionic-native",
-  "version": "4.7.1-beta.1",
+  "version": "4.7.0",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "ionic-native",
-  "version": "4.7.0",
+  "version": "4.7.1-beta.1",
```
danielsogl: Please remove this change

frankkoenigstein: done

PR #2572, opened 2018-06-28
Wrapping new apis in cordova-plugin-ble-central

dodoinblue:


src/@ionic-native/plugins/ble/index.ts
```diff
@@ -579,4 +579,41 @@ export class BLE extends IonicNativePlugin {
...
+  @Cordova()
+  connectedPeripheralsWithServices (services: string[]): Promise<any[]> {
+    return;
+  }
+
+  /**
+   * Find the connected peripherals offering the listed service UUIDs.
+   * iOS only
+   *
+   * @param {string[]} services List of peripheral UUIDs
```
danielsogl: rename services to uuids

dodoinblue: Ah. Good catch. Thank you.

PR #2582, opened 2018-07-04
fix(app-version): fix getVersionCode to return number

tiangolo:
fix(app-version): fix getVersionCode to return a `number` Promise as is in the actual implementation instead of a `string`. Also update and clarify the method docs.

### DIFF
--- a/src/@ionic-native/plugins/app-version/index.ts
+++ b/src/@ionic-native/plugins/app-version/index.ts
@@ -51,11 +51,13 @@ export class AppVersion extends IonicNativePlugin {
   getPackageName(): Promise<string> { return; }
 
   /**
-   * Returns the build identifier of the app, e.g.: 10203
-   * @returns {Promise<number>}
+   * Returns the build identifier of the app.
+   * In iOS a string with the build version like "1.6095"
+   * In Android a number generated from the version string, like 10203 for version "1.2.3"
+   * @returns {Promise<string | number>}
    */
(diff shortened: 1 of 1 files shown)


src/@ionic-native/plugins/app-version/index.ts
```diff
@@ -37,28 +37,28 @@ import { Cordova, Plugin, IonicNativePlugin } from '@ionic-native/core';
...
 
   /**
-   * Returns the build identifier of the app
-   * @returns {Promise<string>}
+   * Returns the build identifier of the app, e.g.: 10203
+   * @returns {Promise<number>}
    */
   @Cordova()
-  getVersionCode(): Promise<string> { return; }
+  getVersionCode(): Promise<number> { return; }
```
danielsogl: That's true for Android but not for iOS.
An iOS version code can look like this "1.2.3.4"

tiangolo: Thanks for the review. You are right.

I just checked the code: https://github.com/whiteoctober/cordova-plugin-app-version/blob/master/src/ios/AppVersion.m

It seems it is returning the `CFBundleVersion`, then I found here in the docs that is seems to be the build version, as you say, as a string: https://developer.apple.com/library/archive/technotes/tn2420/_index.html

I just updated it and the doc string accordingly.

tiangolo:
Thanks! :tada:

PR #2633, opened 2018-08-08
feat(local-notifications): add new foreground property

rupendraa:
Update index.ts add new property foreground


danielsogl:
Please fix the circle errors

PR #2648, opened 2018-08-20
update bluetooth-serial plugin for rxjs 6

yelhouti:


danielsogl:
If you have to use rxjs6 you need to install the @beta native packages (ionic-native v5).

Issue #2653, opened 2018-08-23
[v5 branch] Building esm and ngx fails when path contains src


paxtonhare:
**I'm submitting a ...**  (check one with "x")
[x ] bug report
[ ] feature request
[ ] support request => Please do not submit support requests here, use one of these channels: https://forum.ionicframework.com/ or https://ionicworldwide.herokuapp.com/

**Current behavior:**
If your path contains src (/Users/myname/src/ionic-native/....) then building for esm and ngx will fail because of eager `p.replace('src', 'dist')` code in `scripts/tasks/build-esm.ts` and `scripts/build/ngx.ts`

**Expected behavior:**
Should be able to put src elsewhere in my path. Either that, or document that I can't have src in my path.

**Steps to reproduce:**
clone code into a path with src in it. /Users/myname/src/ionic-native
run `npm run build`

I'll send in a PR shortly.

Issue #2652, opened 2018-08-23
[v5 branch] npm run build is failing


paxtonhare:
**I'm submitting a ...**  (check one with "x")
[x ] bug report
[ ] feature request
[ ] support request => Please do not submit support requests here, use one of these channels: https://forum.ionicframework.com/ or https://ionicworldwide.herokuapp.com/

**Current behavior:**
`npm run build` is failing.

```
ionic-native/scripts/build/helpers.ts:36
  properties.forEach(prop => {
             ^
Error: Unexpected property value type << helpers.ts >>
    at ionic-native/scripts/build/helpers.ts:63:15
    at Array.forEach (<anonymous>)
    at Object.getDecoratorArgs (ionic-native/scripts/build/helpers.ts:36:14)
    at Object.transformMethod (ionic-native/scripts/build/transformers/methods.ts:10:21)
    at ionic-native/scripts/build/transformers/members.ts:14:16
    at Array.map (<anonymous>)
    at Object.transformMembers (ionic-native/scripts/build/transformers/members.ts:8:31)
    at transformClass (ionic-native/scripts/build/transformers/plugin-class.ts:44:9)
    at ionic-native/scripts/build/transformers/plugin-class.ts:64:14
    at visitNodes (ionic-native/node_modules/typescript/lib/typescript.js:56887:48)
```

**Steps to reproduce:**
clone v5 branch
run `npm run build`

**Other information:**
Looks like this commit https://github.com/ionic-team/ionic-native/commit/f2728f26fe93d07e77c67713fed62be3785fa47a broke it by introducing unknown annotations to the network plugin.

```typescript
  @Cordova({
    eventObservable: true,
    event: 'online',
    element: document  // This is the offending code
  })
```

PR #2654, opened 2018-08-23
fixing issue when there are additional src in the path: /Users/myname…

paxtonhare:
…/src/ionic-native

fixes issue #2653


paxtonhare:
circleci failed because of #2652

PR #2655, opened 2018-08-23
objectInstance properties are not being set correctly. 

paxtonhare:
The check in the `instancePropertySet` function is too aggressive.

This manifests in many plugins just not quite working right because the underlying js objects are not getting all the properties they should be getting.


paxtonhare:
circleci failed because of #2652

PR #2660, opened 2018-08-26
fix wrapEventObservable

nicolashenry:
I almost sure (after some debugging) that it should fix a problem with the latest beta (5.0.0-beta.15) due to this change which have been merged : https://github.com/ionic-team/ionic-native/pull/2622 
My problem is that Network.onConnect() throws an error with the latest v5 (but worked with 5.0.0-beta.14). The error is "RangeError: Maximum call stack size exceeded" because rxjs try to subscribe to the "document" string instead of document object. 
I think it should be easy to reproduce.


ihadeed:
Good catch ... we were calling `get(window, element)` but not assigning the value to `element`. 

Thanks

PR #2679, opened 2018-08-30
fix(contacts): fixing builds

perrygovier:
`checkAvailability()` is called in the contacts plugin without being imported and it is breaking docs builds


danielsogl:
Thanks!

PR #2688, opened 2018-09-03
fix(in-app-review): Transfer Plugin to the correct folder

paulstelzer:
In-App-Review Plugin was in the wrong folder


paulstelzer:
The failed check was not my fault - it was at another plugin ;)

PR #2697, opened 2018-09-04
feat(mixpanel): Update/add functions

SpellChucker:
- `alias` does not require an `originalId`.
- `identify` accepts a `usePeople` parameter for the iOS SDK.
- New methods:
`append`
`deleteUser`
`trackCharge`
`unset`
`union`


SpellChucker:
Not sure why tests are failing, doesn't seem to be related to changes.

paulstelzer:
@SpellChucker No, it's not you - had the same problems because their test not install cordova-plugin-my-plugin correctly ;) - so just ignore it

PR #2711, opened 2018-09-11
feat(janalytics): add  plugin

alex-0407:
add janalytics native plugin


danielsogl:
I will merge your PR but I would appreciate if you could add the jsdoc in a separate PR.

PR #2733, opened 2018-09-22
fix(bluetooth-le): Removed a wrong cordova decorator param

somq:
getAdapterInfo is a promise and had {observable: true} param


somq:
read and subscribe methods was returning ```{ result: <OperationResult> }``` which was wrong.
They now directly return the ```<OperationResult>``` object instead.

PR #2736, opened 2018-09-24
chore(imports): fixing builds

perrygovier:
Ionic Native and RXJS made some changes that impact how RXJS is imported and Cordova decorators are called. 

This fixed the build.


danielsogl:
You are targeting the master branch that uses rxjs 5 and Angular 5. On v5 everything is fine.

PR #2743, opened 2018-09-25
fix(keyboard): use cordova-plugin-ionic-keyboard

### LINKED ISSUE #2306: The keyboard module doesn't seem to work with cordova-plugin-ionic-keyboard
**I'm submitting a ...** 
[X] bug report
[ ] feature request
[ ] support request

**Current behavior:**
The module seems to be targeting ionic-plugin-keyboard which is deprecated, and doesn't seem to be working properly with the cordova-plugin-ionic-keyboard plugin.
For example, the onKeyboardHide() observable is never updated.

kensodemann:
Previously we were using the deprecated keyboard plugin.

Fixes #2306

### DIFF
--- a/src/@ionic-native/plugins/keyboard/index.ts
+++ b/src/@ionic-native/plugins/keyboard/index.ts
@@ -7,7 +7,7 @@ import { Observable } from 'rxjs';
  * @description
  * Keyboard plugin for Cordova.
  *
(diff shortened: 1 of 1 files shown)


anshjn:
Could ionic plugin creators can solve ionic plugin of keyboard?

PR #2744, opened 2018-09-26
feat(in-app-browser): add missing property comments

de-dan:
Property comments are omitted during build time, if they are not marked with double asterisks. Added missing asterisks.


de-dan:
Updated

PR #2751, opened 2018-09-28
fix(keyboard) isVisible property missing

sfaizanh:


kensodemann:
Just a note that we will need to make sure this gets into `v5` as well as `master`. FWIW, this LGTM.

sfaizanh:
Hi @kensodemann should i create a separate PR for v5?

danielsogl:
Nope. I merge master into v5 after every new release.

PR #2758, opened 2018-10-03
docs(geolocation): use more generic / professional message

ajcrites:
This is pretty nitpicky, but I think the existing documentation message is somewhat unprofessional. It's also very plausible that people will use this message in the docs as-is without thinking or realizing that it's intended to be changed (or reading it at all).

I propose a message that is hopefully generic enough to be used by any app if worse comes to worst while still communicating the same message.


mlynch:
Thanks, agreed and not sure how that got in there.

PR #2782, opened 2018-10-18
feat(line-login): add plugin

nrikiji:


src/@ionic-native/plugins/line-login/index.ts
```diff
@@ -0,0 +1,95 @@
...
+})
+@Injectable()
+export class LineLogin extends IonicNativePlugin {
+  /**
+   * Initialize
+   * @param param {any} {channel_id: "your_chanel_id"}
+   * @return {Promise<any>}
+   */
+  @Cordova()
+  initialize(param: any): Promise<any> {
```
danielsogl: Add typings

danielsogl:
Because you are the maintainer of the plugin you should be able to add typing information to the return types.

PR #2790, opened 2018-10-23
[4x] Upgrade of rxjs to 6.x

pedrofurtado:


dwieeb:
I don't think we can update this without it being a breaking change. @pedrofurtado see my comment here: https://github.com/ionic-team/starters/pull/525#issuecomment-432708960

danielsogl:
Use ionic native v5 to use rxjs 6 or install the rxjs-compat backward compatible package.

PR #2805, opened 2018-11-06
docs(sqlite): update usage example

surajrao:
executeSql takes array param not object according to the function declaration in code.


surajrao:
Related to https://stackoverflow.com/questions/53167125/cannot-create-table-sqlite-ionic-3

PR #2825, opened 2018-11-27
fix(local-notifications): update type of the every property

terkill:
The `every` property of `ILocalNotificationTrigger` can be either a string or an object.
This PR updates the type to reflect this.


varunneelakantan:
Please merge this change at the earliest, Thanks!

PR #2917, opened 2019-02-05
feat(magnetometer): add new plugin

afshawnlotfi:
Added support for magnetometer (compass) raw x,y,z readings


afshawnlotfi:
Saw this error in the lint but it was your code so wasn't sure if I had to fix it

PR #2940, opened 2019-02-27
feat(purchases): add new plugin for purchases

vegaro:
Adding Plugin for https://github.com/RevenueCat/cordova-plugin-purchases to use with https://www.revenuecat.com/


vegaro:
linter error fixed in https://github.com/ionic-team/ionic-native/pull/2941

selected-pixel-jameson:
Is there a particular reason this hasn't been merged? Would have been nice to have about 20 days ago considering the fact that both of the other plugins for in app purchases are out of date and poorly documented.

PR #2949, opened 2019-03-06
feat(google-play-games-services): add getPlayerScore function

arzanbhorr:
Add getPlayerScore function to the google-play-games-services plugin, as the function is already available on the cordova plugin: cordova-plugin-play-games-services.


arzanbhorr:
Linter error is fixed by https://github.com/ionic-team/ionic-native/pull/2941

Issue #2990, opened 2019-04-09
Initialize() method not working in @ionic-native/firebase-crashlytics


tuxicoman:
**I'm submitting a ...**  (check one with "x")
[ x ] bug report
[ ] feature request
[ ] support request => Please do not submit support requests here, use one of these channels: https://forum.ionicframework.com/ or https://ionicworldwide.herokuapp.com/

**Current behavior:**

firebaseCrashlytics.initialize() doesn't return the crashlytics instance.
1. it has void type
2. function names are not aligned; initiali**z**e vs initiali**s**e

> Native: tried calling FirebaseCrashlytics.initialize, but the FirebaseCrashlytics plugin is not installed
> Install the FirebaseCrashlytics plugin: 'ionic cordova plugin add cordova-plugin-firebase-crashlytics'

**Expected behavior:**
It should return the crashlytics instance.

**Steps to reproduce:**

Install firebase-crashlytics and call as per describe in the [ionicframework doc](https://ionicframework.com/docs/native/firebase-crashlytics) 

```
import { FirebaseCrashlytics } from '@ionic-native/firebase-crashlytics/ngx';


constructor(private firebaseCrashlytics: FirebaseCrashlytics) { }

...


const crashlytics = this.firebaseCrashlytics.initialize();
crashlytics.logException('my caught exception');
```

**Related code:**
https://github.com/ionic-team/ionic-native/blob/4859387925b1dfeb416b80856fbfb991b6884348/src/%40ionic-native/plugins/firebase-crashlytics/index.ts#L43
https://github.com/ReallySmallSoftware/cordova-plugin-firebase-crashlytics/blob/0a3a15b34d612bece06863d2e8a0dee3e1495ab0/www/crashlytics.js#L43
```
module.exports = {
  initialise: function() {
    return new Crashlytics();
  }
};
```

**Other information:**

**Ionic info:** (run `ionic info` from a terminal/cmd prompt and paste output below):

```
Ionic:

   ionic (Ionic CLI)             : 4.12.0 (/usr/lib/node_modules/ionic)
   Ionic Framework               : @ionic/angular 4.2.0
   @angular-devkit/build-angular : 0.12.4
   @angular-devkit/schematics    : 7.3.8
   @angular/cli                  : 7.3.5
   @ionic/angular-toolkit        : 1.3.0

Cordova:

   cordova (Cordova CLI) : 8.1.2 (cordova-lib@8.1.1)
[...]

PR #2991, opened 2019-04-09
fix(firebase-crashlytics): update initialise method (#2990)

tuxicoman:


src/@ionic-native/plugins/firebase-crashlytics/index.ts
```diff
@@ -40,7 +40,7 @@ export class FirebaseCrashlytics extends IonicNativePlugin {
   @Cordova({
     sync: true
   })
-  initialize(): void {
+  initialise(): any {
```
danielsogl: Can you provide more informations why you have to change the return type? Does the function return something?

tuxicoman: Sure thing: `https://github.com/ReallySmallSoftware/cordova-plugin-firebase-crashlytics#how-to-use-it`

```
var crashlytics = FirebaseCrashlytics.initialise();
crashlytics.logException("my caught exception");
```

PR #3018, opened 2019-05-14
feat(camera-preview): add wrapper for camera preview snapshot

max-in-to:
add new method takeSnapshot detailed in plugin and referenced by this open issue: https://github.com/ionic-team/ionic-native/issues/3004


JoeMeeks:
Is there an @ionic-native/camera-preview@4.20.1 scheduled with this fix?

PR #3025, opened 2019-05-16
feat(biometric-wrapper): add plugin

pranav9k:
Used for a cordova plugin. The plugin captures the biometric data(iris and fingerprint).
The plugin is in the ionic source project itself.


pranav9k:
Please merge this. Let me know if there is any issue.

PR #3028, opened 2019-05-20
feat(cordova-plugin-android-fcm): add plugin

ashwindmk:


danielsogl:
Can you describe the difference between your plugin and the existing ones? Because im not willing to merge another FCM plugin into our wrapper collection for no reason.

ashwindmk:
> Can you describe the difference between your plugin and the existing ones? Because im not willing to merge another FCM plugin into our wrapper collection for no reason.

This cordova plugin is a general purpose Android FCM framework where anyone can write custom native Java code for building notifications/sending payload to one/multiple third-party Android SDKs even when app is in background. Also I required a method updateToken which does not take any parameter and simply updates the token to server through native Java code. The existing cordova plugin does not allow to modify notification when received in background.

ashwindmk:
Ok, there is no need to merge this plugin. I came up with an alternate solution, there is no need for any wrapper for my cordova-plugin. Closing the PR.

PR #3037, opened 2019-05-27
feat(in-app-purchase-2): update TypeScript definitions

j3k0:
Hey, I'm the maintainer of `cordova-plugin-purchase`, the cordova plugin used by `in-app-purchase-2`. I made some changes requested by users to the TypeScript definition, but I'm not ionic developer, I'm new to TypeScript, and don't know how to test the changes ;-)

Looks simple and it should probably work, but it'd be better to actually try it!

On a different note: how to update the page https://ionicframework.com/docs/native/in-app-purchase-2 ? The plugin now supports macOS and Microsoft UWP, plus it'd be nice to show more documentation, I can probably help with that.

### DIFF
--- a/src/@ionic-native/plugins/in-app-purchase-2/index.ts
+++ b/src/@ionic-native/plugins/in-app-purchase-2/index.ts
@@ -4,7 +4,7 @@ import { Injectable } from '@angular/core';
 /**
  * @name In App Purchase 2
  * @description
- * In-App Purchase for Cordova on iOS, Android and Windows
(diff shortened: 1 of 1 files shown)


paulstelzer:
@j3k0 thanks for your PR! I tested with new version 5.7.0 of ionic native but I have a problem with it :)

All your properties are set as REQUIRED. So is it really necessary to add every property? If not (and I think it's not required), make the properties optional. For example:

`lastRenewalDate?: Date;`

It's easy by adding the question mark (https://www.typescriptlang.org/docs/handbook/interfaces.html)

PR #3044, opened 2019-05-28
feat(ocr): add plugin

rapropos:
New wrapper for plugin https://github.com/NeutrinosPlatform/cordova-plugin-mobile-ocr.

### DIFF
--- a/src/@ionic-native/plugins/ocr/index.ts
+++ b/src/@ionic-native/plugins/ocr/index.ts
@@ -1,0 +1,167 @@
+import { Injectable } from '@angular/core';
+import { Cordova, IonicNativePlugin, Plugin } from '@ionic-native/core';
+
+export enum OCRSourceType {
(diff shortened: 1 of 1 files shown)


kensodemann:
Question:

The docs for this say to import `MobileOCR`: https://github.com/ionic-team/ionic-native/blob/master/src/@ionic-native/plugins/ocr/index.ts#L125

But the export is just `OCR`: https://github.com/ionic-team/ionic-native/blob/master/src/@ionic-native/plugins/ocr/index.ts#L154

Which one did you intend to be correct?

danielsogl:
@kensodemann `OCR` should be the correct export.

kensodemann:
@danielsogl it looks like you fixed part of the docs but missed a couple of spots. I created a PR for the rest of it: https://github.com/ionic-team/ionic-native/pull/3158

PR #3068, opened 2019-06-19
refactor(core): allow rxjs 5.5 as peer dependency

ltm:
This would allow Ionic v3 apps to use Ionic Native v5 wrappers.

### DIFF
--- a/scripts/tasks/publish.ts
+++ b/scripts/tasks/publish.ts
@@ -30,7 +30,7 @@ const DIST = path.resolve(ROOT, 'dist/@ionic-native');
 const PACKAGES = [];
 
 const MIN_CORE_VERSION = '^5.1.0';
-const RXJS_VERSION = '^6.5.0';
+const RXJS_VERSION = '^5.5.0 || ^6.5.0';
 
 const PLUGIN_PEER_DEPENDENCIES = {
   '@ionic-native/core': MIN_CORE_VERSION,


danielsogl:
Ionic-native 5.x is not compatible with ionic 3.

nphyatt:
👋 @danielsogl We're looking to make it so that Ionic 3 & Ionic Native v5 actually do end up being compatible. We think it'd create a better experience for our users both open source and commercial if we could get compatibility between them. Any insight you could give us into the technical reasons that make this impossible? Seems like the main thing is a difference between the `rxjs` versions. Any insight you have into the issue would be much appreciated.

danielsogl:
Hey @nphyatt.
Because v5 is build with Angular 7, RxJS 6 and TypeScript 3.2, the packages are not compatible with Ionic v3.

If Ionic V3 users want to use Native V5 they should follow this guide here: https://github.com/ionic-team/ionic-native#es2015typescript

ltm:
@danielsogl As far as I can tell, Ionic Native v5 wrappers do work in an Ionic 3 app despite the difference in major versions of those dependencies. I've created a basic app to illustrate this at ltm/ionic3-native5@04ba856fbbf0abc6ec2c47b3b8f92ace4b0ccf47. RxJS 5, for instance, exports `Observable` under both `rxjs/Observable` and `rxjs` (like RxJS 6).

Since there's a considerable number of Ionic 3 apps, I believe it would benefit both developers and Ionic Native maintainers to maintain a single version of Ionic Native supporting both Ionic 3 and Ionic 4.

PR #3108, opened 2019-07-21
doc(unvired-cordova-sdk): improve doc

srinidhirao:


srinidhirao:
I don't see the docs for my plugin in Ionic Native site. Do you have an idea of the process involved to get the docs published?

PR #3122, opened 2019-07-31
feat(cordova-plugin-firebasex): v3 Add plugin

Swamp3:
Added the firebasex plugin to ionic-native v3


danielsogl:
We don't support the V3 Branch anymore. But it is possible to use Ionic-Native V4 with Ionic V3

PR #3146, opened 2019-08-20
fix(network): rename onchange method name

ziyaddin:


N0ps32:
Why was this included in a minor release? This looks like a breaking change. Can I safely upgrade to 5.13.0 without breaking my code that relies on .onchange existing?

PR #3164, opened 2019-09-11
feat(sign-in-with-apple): add plugin

keiya:
this is an ionic-native wrapper for '[cordova-plugin-sign-in-with-apple](https://github.com/twogate/cordova-plugin-sign-in-with-apple)' that is for apple's new authentication architecture 'Sign in with Apple'.

to use the plugin:
```bash
ionic cordova plugin add  https://github.com/twogate/cordova-plugin-sign-in-w
ith-apple.git
```

you should need XCode >= 11.0 beta 6 and iOS (Simulator) >= 13 to run this plugin.

blocks: https://github.com/ionic-team/ionic-native/pull/3165


danielsogl:
As far as I'm concerned, I do not want to merge a wrapper for Sing In With Apple unless I can assure myself that the repositories are properly maintained. Likewise, I do not want to merge more than one wrapper because the functionality is quite simple and the plugins should not be very different.

PR #3165, opened 2019-09-12
add signin-with-apple

nrikiji:
https://github.com/nrikiji/cordova-signin-with-apple-plugin
I make this plugin so I made it compatible with ionic-native

block: #3164


danielsogl:
As far as I'm concerned, I do not want to merge a wrapper for Sing In With Apple unless I can assure myself that the repositories are properly maintained. Likewise, I do not want to merge more than one wrapper because the functionality is quite simple and the plugins should not be very different.

PR #3173, opened 2019-09-20
fix(file): method writeExistingFile should return same type as writeFile

5im0n:
As the `writeExistingFile` method use `writeFile` method is should return the same type.


5im0n:
Hey @danielsogl,

I made this pull request for that Typescript don't throw error when I want to use the result of the `writeExistingFile` method.

With the `writeFile` method:
```typescript
this.file.writeFile('path', 'fileName', 'data', { replace: true }).then(fileEntry => {
   console.log(fileEntry.nativeURL); // Possible cause fileEntry is type any
}):
```
With the `writeExistingFile` method:
```typescript
this.file.writeExistingFile('path', 'fileName', 'data').then(fileEntry => {
   console.log(fileEntry.nativeURL); // Impossible cause fileEntry is type void
}):
```

As the method `writeExistingFile` use `writeFile`, the return type should be `any` as `writeFile`.

PR #3181, opened 2019-09-27
feat(emm-app-config): add plugin

5im0n:
Add [emm-app-config](https://github.com/oracle/cordova-plugin-emm-app-config) plugin to retrieve EMM application configuration


paulosllvn (commented):
Where is this file?

PR #3184, opened 2019-09-29
fix(web-server): add path to response

### LINKED ISSUE #3126: Ionic Webserver wrong cordova plugin install command and wrapper missing path attribute in response
**I'm submitting a ...**  (check one with "x")
[x] bug report
[ ] feature request
[ ] support request => Please do not submit support requests here, use one of these channels: https://forum.ionicframework.com/ or https://ionicworldwide.herokuapp.com/

**Current behavior:**
Ionic native does not correctly implement the Response in `@ionic-native/web-server/ngx `
Path attribute is supported by `https://github.com/bykof/cordova-plugin-webserver` but not in the Ionic wrapper.
**Expected behavior:**
Path attribut should be supported

**Steps to reproduce:**
Install `https://ionicframework.com/docs/native/web-server` and try use path attribute.

myin142:
Partially fixes #3126. It does not fix the wrong npm entry.

I have added optional body and path to the response type. Since path is supported by the cordova plugin.

### DIFF
--- a/src/@ionic-native/plugins/web-server/index.ts
+++ b/src/@ionic-native/plugins/web-server/index.ts
@@ -4,7 +4,8 @@ import { Observable } from 'rxjs';
 
 export interface Response {
   status: number;
-  body: string;
+  body?: string;
+  path?: string;
   headers: { [key: string]: string};
 }
 


JohnRSim:
@myin142 with capacitor 7 android 15 
I'm getting 403 authentication issue when passing path to an image held in the app. 

```
{
  status: 200,
  path: '/assets/sample_artwork.png',
  headers: {
     'Content-Type':   'image/png',
  }
}
```

This works fine - 
```
{
  status: 200,
  body: '<html>Hello World2</html>',
  headers: {
	  'Content-Type': 'text/html'
  }
}
```

myin142:
@JohnRSim I would suggest opening a new issue for this as this PR seems to just extend the Response interface and doesn't actually handle anything. Since this was a long time ago, I also don't even know what this was for so I can't help you with it.

JohnRSim:
This can be closed I can access files directly from the store I was trying to share files from static path in the app

PR #3189, opened 2019-10-04
feat(secure-store-echo): setted newest plugin wrapper

juanmoide:
This feature adds new secure storage wrapper that use other cordova plugin instead of current. This plugin fix compatibilities with android new version (Q/10).


juanmoide:
I don´t know if this new plugin will be merged with current, but having the wrapper meanwhile plugin it´s not merged will save problems with those compabilities that I mentioned before.

Greetings!

PR #3190, opened 2019-10-04
feat(secure-storage-echo): added wrapper to newest plugin

juanmoide:
In some projects I am working on, there are certain incompatibilities with Android Q/10 that It is fixed bt another Cordova plugin. I have created a new wrapper for that plugin can be used and we can start our Apps without problems. I don't know if this plugin is merged with the current one, so for security and necessity for the majority of developers who use this plugin, I think it is a good opportunity to contribute my grain of sand with this wrapper 😆.

Greetings!

EDIT: The current plugin is no longer mantained


juanmoide:
Related with #3189

PR #3206, opened 2019-10-21
feat(fcm): add hasPermission method 

AlexisCaffa:
[Show readme.md](https://github.com/andrehtissot/cordova-plugin-fcm-with-dependecy-updated#version-320-16092019)  

#### Checking for permissions
Useful for IOS. On android, it will always return `true`.

```javascript
FCMPlugin.hasPermission(function(doesIt){
    // doesIt === true => yes, push was allowed
    // doesIt === false => nope, push will not be available
    // doesIt === null => still not answered, recommended checking again later
    if(doesIt) {
        haveFun();
    }
});
```

### DIFF
--- a/src/@ionic-native/plugins/fcm/index.ts
+++ b/src/@ionic-native/plugins/fcm/index.ts
@@ -71,6 +71,19 @@ export class FCM extends IonicNativePlugin {
   getToken(): Promise<string> {
     return;
   }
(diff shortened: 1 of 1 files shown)


kytwb:
Closed via https://github.com/ionic-team/ionic-native/pull/3331.

PR #3226, opened 2019-11-11
fix(firebase-x): update types to match cordova-plugin-firebasex 7.0.0

SnaiCrys:
* (Android) *BREAKING CHANGE* Rework `verifyPhoneNumber()`
    * Remove redundant `verified` in returned credentials object.
    * Support mocking of instant verification for `verifyPhoneNumber()` on Android for integration testing.
* (iOS) *BREAKING CHANGE* Rework `verifyPhoneNumber()` 
    * Return the same credential object structure as Android.
*  Add `signInWithCredential()` to sign user into Firebase account and `linkUserWithCredential()` to link user account with credentials obtained via `verifyPhoneNumber()`.
* (iOS) Add `getAPNSToken()` plugin API method to get the APNS token on iOS.
* (iOS) Add `onApnsTokenReceived()` to register a callback function to be invoked when the APNS token is allocated.

### DIFF
--- a/src/@ionic-native/plugins/firebase-x/index.ts
+++ b/src/@ionic-native/plugins/firebase-x/index.ts
@@ -367,63 +367,74 @@ export class FirebaseX extends IonicNativePlugin {
   }
 
   /**
-   * Request a verification ID and send a SMS with a verification code. Use them to construct a credential to sign in the user (in your app).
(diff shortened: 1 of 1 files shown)


l3ender:
Given that there are breaking changes, this will probably need to be in the next Ionic major release. Could anyone on the Ionic team comment on when this could be included and released?

Issue #3152, opened 2019-08-29
'beforeload' event of in-app-browser not working, becouse no callback argument.


nobu222:
**I'm submitting a ...**  (check one with "x")
[x] bug report
[ ] feature request
[ ] support request => Please do not submit support requests here, use one of these channels: https://forum.ionicframework.com/ or https://ionicworldwide.herokuapp.com/

**Current behavior:**
InAppBrowser cordova plugin emit event, callback function has only one argument< InAppBrowserEvent>.
`(params: InAppBrowserEvent) => void`

But, 'beforeload' event callback function has 2 arguments, 
`(params: InAppBrowserEvent, callback: (url:string) => void) => void`

Now, ionic-native in-app-browser plugin provide only one argument, when 'beforeload' too.
Open url depends on called callback function<second argument>.

**Expected behavior:**
When 'beforeload' event provide all arguments, InAppBrowserEvent and Callback Function.

**Related code:**
```
const browser = this.inAppBrowser.create(url, '_blank', this.options);
browser.on('beforeload').subscribe((...args) => {
        console.log(args); // args is only InAppBrowserEvent
      });
```

Case of Cordova Plugin.
```
inAppBrowserRef.addEventListener('beforeload', beforeloadCallBack);

function beforeloadCallBack(params, callback) {
    if (params.url.startsWith("http://www.example.com/")) {
        // Load this URL in the inAppBrowser.
        callback(params.url);
    } else {
        // The callback is not invoked, so the page will not be loaded.
        $('#status-message').text("This browser only opens pages on http://www.example.com/");
    }
}
```

How can I provide two argument in ionic-native plugin?
If it possible, I fixed and pull request this issue.
thx.

ebk46:
This is an issue I am having as well.

Calinior:
Was this solved by any chance?

kgkg:
Is there any way to use this event in Ionic 4 with new on('beforeload').subscribe interface? Right now I see no way of calling callback, so InAppBrowser hangs on processing this event.

mread1208:
Currently facing this issue as well.  How do we pass the param.url back to the InAppBrowser so it will navigate as expected?

mread1208:
Submited a PR to add the "_loadAfterBeforeload" method to the InAppBrowser instance, I was able to get this working with `"@ionic-native/in-app-browser": "^4.20.0"` since we're still on Ionic v3.

```javascript
const iabRef = this.iab.create(url, target, options);
[...]

PR #3249, opened 2019-12-03
feat(in-app-browser): add "beforeload" callback method

mread1208:
Addressing issue #3152 for version 4 of this plugin.


mread1208:
Since we're still using Ionic v3, we're using `"@ionic-native/in-app-browser": "^4.20.0"` of this plugin.  Is it possible to merge into the `v4` branch or do I need to make this fix in `master` and just point to my forked version for the time being until we update our app to Ionic v4?

PR #3263, opened 2019-12-20
feat(firebase-analytics): add resetAnalyticsData to firebase-analytics

manniniandrea:
added resetAnalyticsData method to firebase-analytics


manniniandrea:
Wrong branch

PR #3293, opened 2020-01-22
fix(interface): add property openUrl & fix Example

osben:


codewise-nicolas:
@CalumMurray care to re-try your pull request as it seems the author of this one has possibly abandoned theirs.

PR #3320, opened 2020-02-19
feat(in-app-purchase-2): update types and add getApplicationUsername function

nvahalik:
This commit does a couple of things:

1. It updates the types of a couple of attributes which are not correctly typed.
2. It updates the return value of `IAPProductEvents` members so that they can be chained.
3. Adds `applicationUsername` which is vital for using the IAP2 module with a validation service.


nvahalik:
I've asked the developer of the IAP2 module to review this PR and he has informed me that he believes there is an issue with my changing of Dates to strings. I am going to double-check this and make appropriate changes.

PR #3347, opened 2020-03-15
fix(email-composer): fix error when no mail apps

nvanselow:
When there are no apps installed that can handle "mail", then the "apps" array is returned as `null` from the plugin. This minor change ensures we don't try to call `length` on `null` but maintains the behavior of the original implementation.

The error appears as:
> Unhandled Rejection (TypeError): null is not an object (evaluating 'apps.length').

I didn't see any corresponding unit tests to update. Please let me know if I missed something. This did correct the issue when I changed the file directly in the node_modules for my project.


nvanselow:
Update: fixed the commit message to match the "Developer" docs. Apologies, missed that section on my first read when looking for code requirements.

nvanselow:
Thank you!

PR #3365, opened 2020-04-03
feat(native-storage): add initWithSuiteName

tatham:
Required if you want to provide access to storage shared between apps in an app group in iOS.
Relevant section of the documentation for the plugin:
https://github.com/TheCocoaProject/cordova-plugin-nativestorage#usage

### DIFF
--- a/src/@ionic-native/plugins/native-storage/index.ts
+++ b/src/@ionic-native/plugins/native-storage/index.ts
@@ -37,12 +37,14 @@ import { Cordova, IonicNativePlugin, Plugin } from '@ionic-native/core';
 @Injectable()
 export class NativeStorage extends IonicNativePlugin {
   /**
-   * Initialises shared storage with the suite name when using app groups
+   * Initialises shared storage with the suite name when using app groups in iOS
    * @param reference {string}
-   * @returns {Promise<any>}
+   * @returns {Promise<void>}
    */
-  @Cordova()
-  initWithSuiteName(reference: string): Promise<any> {
+  @Cordova({
+    platforms: ['iOS']
+  })
+  initWithSuiteName(reference: string): Promise<void> {
     return;
   }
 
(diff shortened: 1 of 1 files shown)


src/@ionic-native/plugins/native-storage/index.ts
```diff
@@ -36,6 +36,16 @@ import { Cordova, IonicNativePlugin, Plugin } from '@ionic-native/core';
 })
 @Injectable()
 export class NativeStorage extends IonicNativePlugin {
+  /**
+   * Initialises shared storage with the suite name when using app groups
+   * @param reference {string}
+   * @returns {Promise<any>}
+   */
+  @Cordova()
+  initWithSuiteName(reference: string): Promise<any> {
```
danielsogl, line 45: Please add an callback interface of set the return type to `Promsie<void>` if the callback is empty

tatham: I don't understand, do you mean that the return type should be changed to Promise<void> or that it needs some logic to check for empty? I've found that the underlying plugin code doesn't recognise an undefined variable (or an empty string) as a null parameter, and so it never returns an error message.

danielsogl: If your promise callback contains Data you should add an interface describing the data like `Promise<MyCallbackInterface>`.

tatham: OK, I was following the style of the other methods, which return `Promise<any>`, but I will change this to `Promise<void>`, as the plugin implementation never returns the error message.
I have two other questions:

- I've realised that I can add `platforms: ['iOS']` to the decorator, as this is iOS only. Should I do that?
- Do I need to raise a separate PR if I want to add the same changes to the v4 branch?

danielsogl: Yes, you should add the platform. Ionic Native 4 is not supported anymore so you don't have to create another PR.

tatham: Great, thanks. Requested changes have been made.

PR #3367, opened 2020-04-03
feat(clovergo): add CloverGo plugin

adityasharma7:
Here is the link to Cordova plugin
https://github.com/hotwax/cordova-plugin-clovergo

The plugin helps to communicate with [Clover Go](https://www.clover.com/shop/clover-go) device and make payment


adityasharma7:
I will create a separate new PR, hence closing

PR #3403, opened 2020-05-04
fix(appsflyer): new methods and callback for in-App events

amit-kremer93:
new methods:
getSDKVersion
registerUninstall
setCurrencyCode

added callback for in-App events (trackEvent)


ab-kily:
> new methods:
> getSDKVersion
> registerUninstall
> setCurrencyCode
> 
> added callback for in-App events (trackEvent)

Hello @amit-kremer93 ! 

@danielsogl just merged PR #3401 which adds registerUninstall method. I think it is better to rewrite your pull request without this method.

PR #3412, opened 2020-05-10
fix(appsflyer): new methods and callback for in-App events

amit-kremer93:
new methods:
getSDKVersion
setCurrencyCode

added callback for in-App events (trackEvent)


danielsogl:
@amit-kremer93 please fix the build/lint issues.

PR #3421, opened 2020-05-15
feat(ios-aswebauthenticationsession): add plugin

philkry:
Wrapper for ASWebAuthenticationSession API for IOS 12/13 to allow web authentication flows with e.g. Auth0


src/@ionic-native/plugins/ios-aswebauthenticationsession-api/index.ts
```diff
@@ -0,0 +1,45 @@
...
+@Injectable()
+export class IosASWebauthenticationSession extends IonicNativePlugin {
+  /**
+   * This function start an authentication flow in ASWebauthenticationSession
+   * @param callbackUrl {string} Callback URL of your App
+   * @param authorizeURL {number} Authorization URL
+   * @return {Promise<any>} Returns a promise that resolves after finishing ASWebauthenticationSession
+   */
+  @Cordova()
+  start(callbackUrl: string, authorizeURL: string): Promise<any> {
```
danielsogl, line 42: Please add an interface instead of `any` as result.

philkry: i changed it to string as the cordova plugin just returns a plain string, thus i don't think an interface is required?

PR #3430, opened 2020-05-21
feat(fcm): add requestPushPermissionIOS and createNotificationChannelAndroid functions

andrehtissot:
Hi, I'm the maintainer of cordova-plugin-fcm-with-dependecy-updated and there are two methods of which the types are not available in @ionic-native/fcm.

I'm not so sure if it's correct. But the definition for the non-ionic methods are available as https://github.com/andrehtissot/cordova-plugin-fcm-with-dependecy-updated/blob/master/src/FCMPlugin.d.ts

### DIFF
--- a/src/@ionic-native/plugins/fcm/index.ts
+++ b/src/@ionic-native/plugins/fcm/index.ts
@@ -16,6 +16,60 @@ export interface NotificationData {
   [name: string]: any;
 }
 
+export interface IRequestPushPermissionIOSOptions {
+  /**
+   * Options exclusive for iOS 9 support
+   */
+  ios9Support?: {
+    /**
+     * How long it will wait for a decision from the user before returning `false`, defaults to 10
+     */
+    timeout?: number;
+    /**
+     * How long between each permission verification, defaults to 0.3
+     */
+    interval?: number;
+  };
+}
+
+export interface IChannelConfiguration {
+  /**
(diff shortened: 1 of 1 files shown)


RichardBoyder:
@dotNetkow Or anyone able to look at this?

RichardBoyder:
@IonicOnCouch  @IonicProSupport  Anyone able to Merge this for us please?

naomi-mastache:
@danielsogl

shprink:
Hello @andrehtissot,
I have installed cordova-plugin-fcm-with-dependecy-updated v7.3.0 (the latest) and the API definition still does not include `requestPushPermissionIOS` or  `createNotificationChannelAndroid`

image

The problem is that the latest ionic-native API has both already.

andrehtissot:
Hi @shprink

I really respect ionic's work on having these API descriptors. 
But, the reality is that, having one repository describing another is very troublesome. Specially if the first has expectations with other dependencies as well.

For this reason and to have the freedom to change the API and have it supported right away for ionic users, the cordova-plugin-fcm-with-dependecy-updated plugin itself has it's methods definitions, even for specific ionic expectations, as described here https://github.com/andrehtissot/cordova-plugin-fcm-with-dependecy-updated#with-ionic

I want to thank for the work of others of this community, specially the maintainers of this repository, but it's much easier for all to keep it on one repository both functional and its description.

PR #3453, opened 2020-06-10
[Docs] Add premier and Capacitor incompatible metadata to plugins

dotNetkow:
Addresses [this issue](https://github.com/ionic-team/ionic-docs/issues/1231). Full background there.

Last year, two new JSON attributes were added to ionic-docs' native.json file (capacitorIncompatible and premierSlug) that do not exist in Ionic Native plugin metadata itself. Both repos became out of sync and following the "official" update process would cause those attributes to disappear, but they need to exist.

As a result, it's been challenging to update the Ionic docs with any changes to community plugins. This PR fixes that, and moving forward we'll be sure to make changes here first, so that both repos don't get out of sync again.

I'm not very familiar with the Ionic Native build process - I don't _think_ these changes break anything, but appreciate a review from someone who knows more.

### DIFF
--- a/scripts/docs-json/index.ts
+++ b/scripts/docs-json/index.ts
@@ -15,6 +15,8 @@ interface Plugin {
   cordovaPlugin: {
     name: string;
   };
+  premierSlug: string;
+  capacitorIncompatible: boolean;
 }
 
 const rootDir = resolve(__dirname, '../..');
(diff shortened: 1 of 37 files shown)


danielsogl:
Looks fine for me but currently the build script is not able to push the changes to the docs. @perrygovier could help us out here.

dotNetkow:
Awesome, thanks Daniel!

Can you elaborate on what the build script issue is? In terms of getting the docs updated, I've been manually copying the generated 'native.json' file over to ionic-docs.... not great, but it works. We can also chat in Slack, probably easier :)

danielsogl:
The old circle CI script committed the changes directly to the ionic sites repo: https://github.com/ionic-team/ionic-native/blob/master/circle.yml

This script isn't working anymore since months. I also switched to GitHub actions as the CI for the repo. if you are familiar with the process of pushing files to the site repo you can edit the GitHub actions file for me. I think that I don't have the correct permissions to create the required GitHub token.

danielsogl:
I added the required scripts to the actions script but the deploy script should be changed. The GitHub Token should be used to publish the changed files. Otherwise we have to setup an SSH key

danielsogl:
We can use this action for example: https://github.com/ad-m/github-push-action

PR #3456, opened 2020-06-12
fix(app-rate): Add missing "openUrl" property to AppRatePreferences object

CalumMurray:


CalumMurray:
Duplicates #3293

CalumMurray:
Re-opened as older PR duplicated by this was abandoned (#3293

PR #3443, opened 2020-05-29
Add missing interface and parameters

CalumMurray:
1. Add missing "fakeVerificationCode" optional parameter from verifyPhoneNumber() API
2. Use Cordova decorator to reverse the success/failure callback order as they're expected as the last parameters
3. Add signOutUser() API prototype

PR #3457, opened 2020-06-12
fix(firebase-x): indicate callback methods to verifyPhoneNumber

vdias38:
indicate callback methods to verifyPhoneNumber and signInWithCredential functions.

Definition for the non-ionic methods are available as
https://github.com/dpa99c/cordova-plugin-firebasex/blob/master/types/index.d.ts#L120

### DIFF
--- a/src/@ionic-native/plugins/firebase-x/index.ts
+++ b/src/@ionic-native/plugins/firebase-x/index.ts
@@ -490,8 +490,6 @@ export class FirebaseX extends IonicNativePlugin {
    * the verification code along with the verification ID. In this case, the user doesn't need to do anything in order for you
    * to sign them in.
    *
-   * @param {function} success - callback function to pass {object} credentials to as an argument
-   * @param {function} error - callback function which will be passed a {string} error message as an argument
    * @param {string} phoneNumber - phone number to verify
    * @param {integer} timeOutDuration - time to wait in seconds before timing out
    * @param {string} fakeVerificationCode - (optional) to test instant verification on Android ,specify a fake verification code to return for whitelisted phone numbers.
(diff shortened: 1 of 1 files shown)


Sampath-Lokuge:
Hi @vdias38 Can we have release using this change?

vdias38:
> Hi @vdias38 Can we have release using this change?

I hope, but should wait owners to merge it, or use my branch

Sampath-Lokuge:
Thanks, @vdias38 I'll wait till the official release.

CalumMurray:
Duplicates #3443

vdias38:
> Partially duplicates #3443

Hi @CalumMurray , it's not exactly the same, I don't understand why you use callbacks as arguments and how/why you use them. 
As far as I understand to use callbacks you can use directly the plugin interface. The ionic-native provides an interface to replace callbacks by Promise or Observable, then verifyPhoneNumber for example, could be called as below:

```ts
this.firebaseX.verifyPhoneNumber(phone, this.timeOutDuration)
.then( credential => {...})
.catch(error => {...})
```

Implementation above is compatible with following declaration I've submitted to ionic-native, but not with yours.
```js
  @Cordova({
    callbackOrder: 'reverse'
  })
  verifyPhoneNumber(
    phoneNumber: string,
    timeOutDuration: number,
    fakeVerificationCode?: string
  ): Promise<any> {
    return;
  }
```

**Make sense or do I miss something?**
Following same logic I've added `signOutUser` on this #3458

CalumMurray:
Hi @vdias38, I think you're right. I had assumed they were removed during some build step perhaps as other had left the success/error callbacks in, so I also left them. Thanks for adding `signOutUser()`. Since yours is more complete, I think this can replace my PR #3443

PR #3458, opened 2020-06-12
feat(firebase-x): add missing methods, incl. authenticateUserWithApple

vdias38:
Definition for the non-ionic methods are available as
https://github.com/dpa99c/cordova-plugin-firebasex/blob/master/types/index.d.ts


alemit:
I'm interested when this is going to be merged & released

Issue #3477, opened 2020-07-02
@ionic-native/onesignal plugin missing setExternalUserId/removeExternalUserId callback param


cristianofromagio:
**I'm submitting a ...**  (check one with "x")
[ ] bug report
[x] feature request
[ ] support request => Please do not submit support requests here, use one of these channels: https://forum.ionicframework.com/ or https://ionicworldwide.herokuapp.com/

**Current behavior:**
Current `setExternalUserId` method accepts only the `externalId` string param and `removeExternalUserId` accepts no param

**Expected behavior:**
`setExternalUserId` method accepts both `externalId` and the optional callback function param and `removeExternalUserId` accept the optional callback function param

**Other information:**
As stated on the One Signal SDK Reference, since Cordova SDK 2.9.0, both [setExternalUserId](https://documentation.onesignal.com/docs/sdk-reference#setexternaluserid-method) and [removeExternalUserId](https://documentation.onesignal.com/docs/sdk-reference#removeexternaluserid-method) accept a function callback as param

Release info from the relevant tags:

- https://github.com/OneSignal/OneSignal-Cordova-SDK/releases/tag/2.9.0
    + https://github.com/OneSignal/OneSignal-Android-SDK/releases/tag/3.13.0
    + https://github.com/OneSignal/OneSignal-iOS-SDK/releases/tag/2.13.0

**Ionic info:** (run `ionic info` from a terminal/cmd prompt and paste output below):

```
Ionic:

   Ionic CLI                     : 6.3.0 (C:\ProgramData\nvm\v12.16.2\node_modules\@ionic\cli)
   Ionic Framework               : @ionic/angular 4.11.7
   @angular-devkit/build-angular : 0.803.25
   @angular-devkit/schematics    : 7.3.9
   @angular/cli                  : 8.3.25
   @ionic/angular-toolkit        : 2.1.2

Cordova:

   Cordova CLI       : 9.0.0 (cordova-lib@9.0.1)
   Cordova Platforms : android 8.1.0
   Cordova Plugins   : cordova-plugin-ionic-keyboard 2.2.0, cordova-plugin-ionic-webview 4.1.3, (and 6 other plugins)

Utility:

   cordova-res (update available: 0.15.1) : 0.11.0
   native-run                             : 1.0.0

System:

   NodeJS : v12.16.2 (C:\Program Files\nodejs\node.exe)
   npm    : 6.14.4
   OS     : Windows 10
```

PR #3479, opened 2020-07-03
feat(onesignal): add missing callback params

cristianofromagio:
Fix issue #3477

I didn't write the callback function type as none of the SDKs seems to do (they are just generic JSONObject types, like seem here on [Android SDK](https://github.com/OneSignal/OneSignal-Android-SDK/blob/68f6ade7ac30e8d3038213957134627940079cb4/OneSignalSDK/onesignal/src/main/java/com/onesignal/OneSignal.java#L204)). On the [SDK Reference](https://documentation.onesignal.com/docs/sdk-reference#setexternaluserid-method) entry example they hint at what the result content might be, so I add this as a comment above the method.


danielsogl:
@cristianofromagio please resolve the conflicts

Issue #3506, opened 2020-08-10
@ionic-native/ble : Sequences read from BLE can be misordered


CodeTroopers:
**I'm submitting a**
[x] bug report
[ ] feature request
[x] support request

**Current behavior:**
Sequences read from BLE can be misordered

**Expected behavior:**
Sequences should be received in the sending order or with a order number to allow reorder the sequences

**Steps to reproduce:**
Sorry the bug is random on an intensive using of BLE reading

**Other information:**
The bug has been fixed in the version 1.2.4 of [cordova-plugin-ble-central](https://github.com/don/cordova-plugin-ble-central) with a new parameter to get the sequence order. The PR is [here](https://github.com/don/cordova-plugin-ble-central/pull/656). 
From IONIC, I receive an Observable on data and I don’t see how access this parameter.

The issue can also be found [on Stackoverflow](https://stackoverflow.com/questions/63264971/how-to-use-the-second-parameter-of-callback-of-cordova-plugin-ble) and [on IONIC forum](https://forum.ionicframework.com/t/ble-how-to-use-sequence-number-in-an-angular-app/193938)

**Ionic info:** 
```
Ionic:

   Ionic CLI                     : 5.4.15 (/usr/local/lib/node_modules/ionic)
   Ionic Framework               : @ionic/angular 4.11.10
   @angular-devkit/build-angular : 0.803.26
   @angular-devkit/schematics    : 8.3.26
   @angular/cli                  : 8.3.26
   @ionic/angular-toolkit        : 2.1.2

Cordova:

   Cordova CLI       : 9.0.0 (cordova-lib@9.0.1)
   Cordova Platforms : android 8.1.0
   Cordova Plugins   : cordova-plugin-ionic-keyboard 2.2.0, cordova-plugin-ionic-webview 4.1.3, (and 13 other plugins)

Utility:

   cordova-res : not installed
   native-run  : 1.0.0

System:

   Android SDK Tools : 26.1.1 (/Users/anonymous/Library/Android/sdk)
   ios-sim           : 8.0.2
   NodeJS            : v12.16.2 (/usr/local/bin/node)
   npm               : 6.14.7
   OS                : macOS Catalina
   Xcode             : Xcode 11.6 Build version 11E708
```

CodeTroopers:
To resume :

[...]

PR #3509, opened 2020-08-14
fix(ble) #3506 : leave a chance to client to reorder sequences 

CodeTroopers:
To resume :

- Cordova doesn't guarantee callback ordering relative to when they are queued. See [this analyze](https://github.com/don/cordova-plugin-ble-central/issues/625) of @timburke. Thanks to him.
- cordova-plugin-ble-central has fixed this by adding a second parameter to the callback to leave the client checking and reordering the sequences order. This fix has been included in version 1.2.4. [Here](https://github.com/don/cordova-plugin-ble-central/pull/656) the PR
- The second parameter of plugin callback was not accessible from IONIC wrapper used to convert callback to observable
- I changed the option of wrapper by adding the destruct option. Now the value returned by subscription on the observable is an array with at index 0 the data value and at index 1 the sequence order. This is a breaking change

### DIFF
--- a/src/@ionic-native/plugins/ble/index.ts
+++ b/src/@ionic-native/plugins/ble/index.ts
@@ -455,6 +455,8 @@ export class BLE extends IonicNativePlugin {
    * @param {string} serviceUUID  UUID of the BLE service
    * @param {string} characteristicUUID  UUID of the BLE characteristic
    * @return {Observable<any>} Returns an Observable that notifies of characteristic changes.
+   * The observer emit an array with data at index 0 and sequence order at index 1.
+   * The sequence order is always undefined on iOS. On android it leave the client to check the sequence order and reorder if needed
    */
   @Cordova({
     observable: true,
--- a/src/@ionic-native/plugins/ble/index.ts
+++ b/src/@ionic-native/plugins/ble/index.ts
@@ -458,6 +458,7 @@ export class BLE extends IonicNativePlugin {
    */
   @Cordova({
     observable: true,
+    destruct: true,
     clearFunction: 'stopNotification',
     clearWithArgs: true,
   })


CodeTroopers:
@danielsogl you have forgotten to mention the breaking change on the changelog and perhaps in the documentation

Naoray:
@CodeTroopers I might have run into this issue of receiving data which now looks different. Could you explain the difference between now and prior in the return values? As far as I understand from your comment in the code the result now contains all data at index `0` of the array and the sequence at index `1`, right?

So would the following code snippet be correct in recieving the data from the notification
```
BLE.startNotification(state.device.id,
      characteristic.serviceUUID,
      characteristic.characteristicUUID
    ).subscribe((arrayBuffer) => {
      console.log('test notification')
      console.log(new Uint8Array(arrayBuffer)[0])
      console.log('test notification end')
      characteristic.parser(arrayBuffer)
    })

**Comment by CodeTroopers** [comment]
Yes data are at index 0 of array result and sequence number is at index 1. We had a lot of misordered sequence on Android 8+
Here an example to reorder if needed
```
let tampon: { [key: number]: Uint8Array } = {};
let currentNumber = 0
BLE.startNotification(
      state.device.id,
      serviceUUID,
      characteristicUUID
    ).subscribe((result) => {
     const data = new Uint8Array(result[0])
     const sequenceNumber: number | undefined = result[1];
     if (sequenceNumber !== undefined && sequenceNumber !== currentNumber) {
           // Bad order : save in tampon
           tampon[sequenceNumber] = data
     } else {
           callback(data)
           currentNumber++
     }
     tryReorder()
    })

private tryReorder(): void {
    while (currentNumber in tampon) {
        try {
            const value = tampon[currentNumber];
            delete this.tampon[currentNumber];
            currentNumber++;

            callback(value);
	} catch (err) {
            console.error("Error when reordering sequence", err);
        }
    }
}
```

PR #3527, opened 2020-09-04
feat(firebase-x): Add missing function reloadCurrentUser()

villeraikko:
Added the missing function reloadCurrentUser() for cordova-plugin-firebasex. This is identical to the already implemented function getCurrentUser(), but instead of loading the user from local memory, it reloads it from Firebase.

https://github.com/dpa99c/cordova-plugin-firebasex#reloadcurrentuser


villeraikko:
Related to this issue https://github.com/ionic-team/ionic-native/issues/3492

mhartington:
Thanks for the PR

PR #3539, opened 2020-09-18
feat(background-upload): add plugin

MaximBelov:


src/@ionic-native/plugins/background-upload/index.ts
```diff
@@ -0,0 +1,112 @@
...
+}
+
+/**
+ * @name BackgroundUpload
+ * @description
+ * This plugin does something
+ *
+ * @usage
+ * ```typescript
+ * import { BackgroundUpload } from '@ionic-native/background-upload/npx';
```
danielsogl, line 83: Change it to `/ngx`

danielsogl (changes requested):

src/@ionic-native/plugins/background-upload/index.ts
MaximBelov: @danielsogl 
done

danielsogl (approved):

PR #3490, opened 2020-07-17
feat(diagnostics): Add support for new iOS14-LocationAccuracy Permission

HansKrywaa:
Issue Number: #3488 

The cordova-diagnostics Plugin added a Function for the new ios14 location accuracy authorization API: https://github.com/dpa99c/cordova-diagnostic-plugin/blob/master/CHANGELOG.md

This is my first PR to **ionic-native**, i hope everything is correct 😊

PR #3540, opened 2020-09-18
fix(diagnostics): locationAccuracyAuthorization Property missing in build

EinfachHans:
Just noticed, that the **locationAccuracyAuthorization** Property from my last PR (#3490) is missing in final build's `index.js`. The only difference to the other Properties was the comma/semicolon, so i changed this. Hope it's correct now?

Sorry for the issue in the last PR...


danielsogl (approved):

ryaa:
The problem still exists in the latest build - see switchToSettings

PR #3563, opened 2020-10-22
feat(AppsFlyer): update to 6.x.x

amit-kremer93:
In AppsFlyer Cordova plugin 6.x.x the api was changed from 'track' to 'log'.
Ionic-Native AppsFlyer plugin support only until version 5.4.30

### DIFF
--- a/src/@ionic-native/plugins/appsflyer/index.ts
+++ b/src/@ionic-native/plugins/appsflyer/index.ts
@@ -94,7 +94,7 @@ export class Appsflyer extends IonicNativePlugin {
    * @param {string} eventName custom event name, is presented in your dashboard
    * @param {AppsflyerEvent} eventValues event details
    */
-  @Cordova({ sync: true })
+  @Cordova()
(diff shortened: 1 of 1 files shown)


jefflnorton:
Also looking to update AppsFlyer Cordova plugin to v6.x to support iOS 14 AppTrackingTransparency changes.

src/@ionic-native/plugins/appsflyer/index.ts
```diff
@@ -95,7 +95,7 @@ export class Appsflyer extends IonicNativePlugin {
    * @param {AppsflyerEvent} eventValues event details
    */
   @Cordova({ sync: true })
```
ihadeed, line 97: This function shouldn't be sync

https://github.com/AppsFlyerSDK/appsflyer-cordova-plugin/blob/master/www/appsflyer.js#L128-L131

ihadeed:
LGTM 

The comment above is about code that's not introduced by this PR

danielsogl (approved):

PR #3565, opened 2020-10-26
docs(cordova-plugin-firebase-analytics): Adding instructions for capacitor support on Android.

tnovau:
Adding instructions for firebase analytics capacitor support on Android, because the plugin require cordova variables.


ericgopak (approved):
Thanks a lot!
I've followed your conversation from `cordova-plugin-frebase-analytics` repo while looking for a fix.
Tested - it works :+1:

tnovau:
> Thanks a lot!
> I've followed your conversation from `cordova-plugin-frebase-analytics` repo while looking for a fix.
> Tested - it works 👍

You're welcome! I think it will be easier for all if this is documented in ionic native. Since it says that capacitor is supported but no how to use it.

danielsogl (approved):

PR #3623, opened 2021-02-11
fix(bluetooth-le): correct BOND and UNBOND response type

NunoSav:
Resolves https://github.com/ionic-team/ionic-native/issues/3622

### DIFF
--- a/src/@ionic-native/plugins/bluetooth-le/index.ts
+++ b/src/@ionic-native/plugins/bluetooth-le/index.ts
@@ -102,6 +102,38 @@ export interface RespondParams {
   offset?: number;
 }
 
+export interface ConnectionParams {
+  /** The address/identifier provided by the scan's return object */
+  address: string;
+  /** Automatically connect as soon as the remote device becomes available (Android) */
+  autoConnect?: boolean;
+  /**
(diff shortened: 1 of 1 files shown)


NunoSav:
I came back to this since i started using the local changes in my project and saw errors. As i went through the BLE file i saw a lot of inconsistencies with the latest published version 🤷‍♂️ .

I followed the [guidelines](https://github.com/ionic-team/ionic-native/blob/master/DEVELOPER.md#testing-your-changes) and tried to fix all inconsistencies. `npm build` ran without issues, i then imported the local changes into my project and had no issues either.

The last step, `gulp lint`, returned [this error](https://stackoverflow.com/questions/55921442/how-to-fix-referenceerror-primordials-is-not-defined-in-node). To fix this i applied this [solution](https://stackoverflow.com/a/58394828/6863813) (under **If you are using npm** header), ran `npm i` and got a bunch of errors so i used yarn (no issues).

Besides fixing the inconsistencies i've also added 2 missing methods.
All changes were in accordance with the [plugin types](https://github.com/randdusing/cordova-plugin-bluetoothle/blob/master/types/index.d.ts).

Hope to hear from you soon 🤞

PR #3682, opened 2021-06-06
feat(approov-advanced-http) add wrapper for approov advanced http cordova plugin

Shivang8941:


danielsogl (changes requested):
You have to rename the class name to something like `ApproovHttp` because there is already a Http plugin

Shivang8941:
> 
> 
> You have to rename the class name to something like `ApproovHttp` because there is already a Http plugin

Shivang8941:
> 
> 
> You have to rename the class name to something like `ApproovHttp` because there is already a Http plugin

We want to extend the same plugin that's why we've kept the same name. This way, it will be easier for existing users to migrate to this plugin without a major refactor.

danielsogl:
@Shivang8941 I can understand your point but we dont accept duplicated plugin names. It is also recommended to switch to capacitor and its plugins to handle Http requests etc.

Shivang8941:
> 
> 
> @Shivang8941 I can understand your point but we dont accept duplicated plugin names. It is also recommended to switch to capacitor and its plugins to handle Http requests etc.

For time-being can we only update the plugin name? Will that suffice?

Shivang8941:
@danielsogl Can you please have a look at this again. I've updated the package name.

danielsogl (approved):

PR #3709, opened 2021-08-09
fix(purchases): remove static for logIn, logOut methods

evanharvey:
Purchases plugin is not working in my ionic project with static method logIn, logOut
that methods will return undefined so I removed static of logIn, logOut methods.

This code must be reviewed by revenuecat developers before marge.


aboedo (approved):
@evanharvey thanks for doing this! it was an oversight on my part 🤦‍♂️

danielsogl (approved):

PR #3733, opened 2021-09-07
(feat)tealium: update tealium ionic wrapper

tamayok:


craigrouse:
@danielsogl Is there any reason why this was closed rather than merged? We have customers wishing to use it in conjunction with the latest version of our Cordova plugin. Thank you!

src/@ionic-native/plugins/tealium/index.ts
```diff
@@ -1,217 +1,367 @@
...
+
+export class TealiumEvent implements TealiumDispatch {
+	public type: string = 'event';
+	constructor(public eventName: string, public dataLayer: Map<string, any>) {}
+	toJson() {
+		let dictionary: any = {};
+		dictionary['type'] = this.type;
+		dictionary['dataLayer'] = {};
+    this.dataLayer.forEach((k, v) => {
+      dictionary['dataLayer'][k] = v;
```
sarnendumukherjee, line 59: Right now with the current code, it is making the key as value and vice versa. 
It should be like this for both Event and View. 
```
this.dataLayer.forEach((value, key) => {
      dictionary['dataLayer'][key] = value;
});
```

danielsogl:
The PR is pointing to the old structure of the project. @tamayok ore someone else should recreate a new one

PR #3736, opened 2021-09-11
Ionic Native Cleanup

mlynch:
This is a work in progress PR to clean up a bunch of unmaintained, not widely used, or low-quality plugins in Ionic Native. This is part of a broader internal effort here at Ionic to audit and improve the quality of plugins in the community, and also to provide a path forward for the Ionic Native project especially given the growth of Capacitor. 

There are still open questions about what that will look like, but to start, I will be removing plugins that are verified to have not been maintained for many years, have few to no monthly installs, are broken, or are generally of low quality. The data I am using to base that decision off is a mixture of github archive status, last commit date, open issue count, stars, and npm download/publish data. Eventually I would like to also compile/build each plugin into a real app to test if they compile or work, but that will have to come later.

This will be a breaking change once this work is done. Open to feedback or appeals for any of the plugins I end up removing.


mlynch:
Okay I feel good about this first pass. We have some more tangible plans about where we want to take this project and I'll be creating some more PRs in the near future and sharing those plans

Issue #3737, opened 2021-09-12
BLE plugin is missing a method


Mr-Mime:
**I'm submitting a ...** 
 [ ] bug report
 [x] feature request
 [ ] support request => Please do not submit support requests here, use one of these channels: https://forum.ionicframework.com/ or https://ionicworldwide.herokuapp.com/

**Current behavior:**
The [BLE](https://ionicframework.com/docs/native/ble plugin) is missing the `requestConnectionPriority()` method.

**Expected behavior:**
The `requestConnectionPriority()` method should be part of the plugin.

**Steps to reproduce:**
Not needed.

**Other information:**
I will provide a PR ASAP.

**Ionic info:**

```
Ionic:

   Ionic CLI                     : 6.13.1 (/usr/lib/node_modules/@ionic/cli)
   Ionic Framework               : @ionic/angular 5.5.2
   @angular-devkit/build-angular : 0.1100.4
   @angular-devkit/schematics    : 11.0.4
   @angular/cli                  : 11.0.4
   @ionic/angular-toolkit        : 3.0.0

Capacitor:

   Capacitor CLI   : 2.4.4
   @capacitor/core : 2.4.4

Utility:

   cordova-res : not installed
   native-run  : not installed

System:

   NodeJS : v16.7.0 (/usr/bin/node)
   npm    : 7.21.0
   OS     : Linux 5.10
```

PR #3738, opened 2021-09-12
fix(ble): add missing method "requestConnectionPriority"

Mr-Mime:
This adds the missing `requestConnectionPriority` method to the ble plugin.

PR is related to #3737.


huiigary:
Hi, can we add `requestConnectionPriority` to the  BLE plugin? It looks like it's missing.
Thanks.

joseitor:
Can we try to add the requestConnectionPriority, or it's not relevant anymore in new Android APIs?

PR #3887, opened 2021-12-02
feat(magnetometer): add magnetometer wrapper

kitcarson88:


danielsogl (approved):

gelinger777:
this wrapper has a problem. Unsubscribe  works on the javascript level from listener, but it does not stop actually underlying sensor readings, which has to happen using .stop command  of the plugin  . On cordova plugin itself when you start watching it will give you back the ID of the watcher, and that is later used  to stop it. I don't know why this mechanism is not working...

PR #3888, opened 2021-12-05
feat(gyroscope): add gyroscope wrapper

kitcarson88:


danielsogl:
Hey, looks like this plugin is not maintained anymore. You can use the capacitor implementation to use the gyro feature: https://capacitorjs.com/docs/apis/motion

PR #4173, opened 2022-05-04
add - interface : DiagnosticPermissionStatus

rbalet:
Add `DiagnosticPermissionStatus` as interface


danielsogl (changes requested):
Whats the reason you moved the plugin specific interface to the decorators? All plugin relavant code and interfaces should not be moved

rbalet:
> Whats the reason you moved the plugin specific interface to the decorators? All plugin relavant code and interfaces should not be moved

To make them available when imported into a project.

I though moving it into the interface file would make sense.

Where should it goes?

PR #4755, opened 2024-04-10
feat(airship): add plugin

MaximBelov:
Package Changes
https://github.com/urbanairship/urbanairship-cordova/blob/main/MIGRATION.md#package-changes

https://docs.airship.com/platform/mobile/setup/sdk/cordova/

Airship Plugin
https://github.com/urbanairship/urbanairship-cordova/pull/423

### DIFF
--- a/src/@awesome-cordova-plugins/plugins/airship/index.ts
+++ b/src/@awesome-cordova-plugins/plugins/airship/index.ts
@@ -742,7 +742,6 @@ export interface ItemBase {
 /**
  * A channel subscription item.
  * @typedef {object} ChannelSubscriptionItem
- * @memberof PreferenceCenter
  * @property {"channel_subscription"} type
  * @property {string} id the item identifier
  * @property {?CommonDisplay} display display information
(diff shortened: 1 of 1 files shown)


src/@awesome-cordova-plugins/plugins/airship/index.ts
```diff
@@ -0,0 +1,1755 @@
...
+@Plugin({
+  pluginName: 'AirShip',
+  plugin: '@ua/cordova-airship',
+  pluginRef: 'Airship',
+  repo: 'https://www.npmjs.com/package/@ua/cordova-airship',
+  install: 'ionic cordova plugin add @ua/cordova-airship',
+  platforms: ['Android', 'iOS'],
+})
+@Injectable()
+export class AirShip extends AwesomeCordovaNativePlugin {
```
rlepinski, line 44: If we are doing a breaking change this should be `Airship` or does that conflict with the cordova name?

rlepinski:
We now have a capacitor plugin, might not need this anymore? https://github.com/urbanairship/capacitor-airship I bet it has the same issues as cordova though. Ill try to patch both today/tomorrow

src/@awesome-cordova-plugins/plugins/airship/index.ts
MaximBelov: changed to `Airship`

danielsogl (approved):

PR #4929, opened 2025-04-08
fix(screenshot): Update import

MaximBelov:
No matching export in "node_modules/@awesome-cordova-plugins/core/index.js" for import "Plugin"


MaximBelov:
@danielsogl 
Please review and merge

